PR #18, opened 2019-12-20
try to fix parser file because of deprecated usage of syntax

sjanssen2:


sjanssen2:
closed in favor of #19

PR #27, opened 2020-05-29
switch to C17 in order to avoid deprecation warnings

sgtkellox:


sjanssen2:
Please don't merge yet. In addition to the tests of this repo, I asked @sgtkellox to also trigger a PR in the fold-grammars repo to source bgap from this branch and execute all fold-grammars specific tests. If they also pass we are relatively certain nothing will break.

sgtkellox:
C++17 does not need you to specify the exception type in a methods signature. All exception throwing methods have a tag noexcept(false) in the signature. Type specifications may still occure in the throw call when a exception causing error is detected.
https://en.cppreference.com/w/cpp/language/except_spec

sjanssen2:
outdated by #29

PR #29, opened 2020-06-03
Migration to c17

sjanssen2:
Text from @sgtkellox :
C++17 does not need you to specify the exception type in a methods signature. All exception throwing methods have a tag noexcept(false) in the signature. Type specifications may still occure in the throw call when a exception causing error is detected.
https://en.cppreference.com/w/cpp/language/except_spec

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,46 @@
 dist: xenial
+os: linux
 language: cpp
 compiler:
-- gcc
-os:
-- linux
+  - gcc
 
 jobs:
   include:
-    - stage: test
-      addons:
-        apt:
-          packages:
-            - flex
-            - bison
-            - make
-            - libboost-all-dev
-            - libgsl-dev
-            - ghc
-      install:
-        - ./configure
-        - make
-        - sudo make install
-      before_script:
-        # install missing ghc/haskell modules
-        - sudo apt-get install cabal-install -y
-        - cabal update
-        - cabal install random
-        # obtain truth data from secondary github repository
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -1,36 +1,46 @@
 dist: xenial
```
sjanssen2, line 1: Changes on this file were introduced by me, basically to let Travis test on Ubuntu versions 16.04 (as before) and on 18.04 (new). The later ships with a more recent g++ version which would actually raise the warnings @sgtkellox was working on. They were not thrown with 16.04, i.e. we couldn't make out the different of @sgtkellox PR.

aclocal.m4
```diff
@@ -1,6 +1,6 @@
-# generated automatically by aclocal 1.14.1 -*- Autoconf -*-
```
sjanssen2, line 1: auto-generated changes - ignore in your review

configure
```diff
@@ -686,6 +686,7 @@ infodir
 docdir
```
sjanssen2, line 686: auto-generated changes - ignore in your review
see below in file developer_notes for more infos

configure.ac
```diff
@@ -62,8 +62,8 @@ GAP_PIC_FLAGS
 
 
 AC_LANG([C++])
-AX_CHECK_COMPILE_FLAG([-std=c++11], [
```
sjanssen2, line 65: we want to migrate from c++11 to c++17 and use this template with autotools to generate the configure script

PR #31, opened 2020-06-07
activate code style tests

sjanssen2:
I noticed that we have no formal code-style checker integrated for bellman's GAP and in fact source code is - let's say - of various quality.
Aiming for an open source program, I found [cpplint](https://github.com/cpplint/cpplint) that checks code for style adherence against a standard developed by Google. One can for sure debate if this style makes sense in every aspects; however in the end I think any code style is better than none. 

Thus, I am currently - locally (`cpplint src/* src/ambiguity_cfg_gen/* src/cfg/* src/expr/* src/printer/* src/specialize_grammar/* src/statement/* src/type/* src/util/*`) - executing cpplint in all source files in src/* and find a LOT of warnings, namely 42,594.

This PR is already massive and almost un-reviewable. I only made changes to the syntax of the code, not touching any functionality. This brings the number of code-sytle warnings down to 912, i.e. a lot is solved but there is still some room for improvement. Actually, the remaining warnings roughly group into:
 - "Is this a non-const reference? If so, make const or use a pointer"
 - "Single-parameter constructors should be marked explicit"
For which I currently don't know how to best handle them. I hope @sgtkellox can find some solutions.

Future directions: we would also need to fix style in `rtlib/` and ultimately also ensuring that the produced code adheres to the same codestyle guide. But these are issues for separate PRs.

Since this PR introduces so many changes, I don't only want to rely on gapc's unit tests. In addition, I also issues a PR in fold-grammars, using gapc from this branch. All fold-grammars tests pass: https://travis-ci.org/github/jlab/fold-grammars/builds/696813680


sjanssen2:
@sgtkellox 
1) I hope this is only due to a misconfiguration on your machine. I had `cpplint` running in Travis (see https://travis-ci.org/github/jlab/gapc/jobs/695891887) but since we have a LOT of codestyle issues to fix first, I deactivated it for now. (`#- cpplint src/* rtlib/*`)

2) because it broke downstream tests with fold-grammars and I still don't know what I am doing with adding this keyword. Thus, I decided to leave as is for now and first clean up all other codestyle issues. Once that is done, we need to come back to it!

Thanks for your review!

sjanssen2:
Many thanks to both of you guys @sgtkellox @kmaibach !

I hope future pull request will be easier to review, but we need some more effort to get the style correct. Thus, please have a look at my other PR #32 which does the same as this one, but for the `rtlib` sub-directory.

PR #32, opened 2020-06-11
Codestyle rtlib

sjanssen2:
Similar to #31 for the sub-directory `rtlib/`
Bringing the number of warnings down from 4805 to "only" 606.
Let us merge #31 first!


rtlib/hash_stats.hh
```diff
@@ -1,5 +1,5 @@
-#ifndef HASH_STATS
-#define HASH_STATS
+#ifndef RTLIB_HASH_STATS_HH_
```
kmaibach, line 24: Why is it considered a better style to add underscores at the end?

rtlib/table.hh
```diff
@@ -447,7 +451,8 @@ namespace Table {
       if (a>b)
         std::swap(a, b);
 
-      //std::cerr << "i " << i << " j " << j << " a " << a << " b " << b << std::endl;
+      // std::cerr << "i " << i << " j " << j << " a " <<
+      // a << " b " << b << std::endl;
```
kmaibach: I'm not sure if I like this style, personally. To split a single-line comment into two lines because - I assume - the comment is considered too long? But that's just personal preference.

rtlib/hash_stats.hh
sjanssen2: honestly: I don't know and I cannot find anything about it in Google's styleguide: https://google.github.io/styleguide/cppguide.html#Header_Files
But hey, as long as it doesn't break anything we can comply to this rule

rtlib/table.hh
sjanssen2: I personally agree, but for (historic) reasons, source code lines should never exceed 80 characters. I guess this is due to the fact that early day's terminals were exactly 80 characters wide and if you think of tools like `diff` it makes sense to have finite numbers of chars in a line. I would prefer a larger cut-off, but if the world picks 80 than let's go with it.

rtlib/bigint.hh
```diff
@@ -22,8 +22,8 @@
 }}} */
 
 
-#ifndef BIGINT_HH 
-#define BIGINT_HH
+#ifndef RTLIB_BIGINT_HH_
+#define RTLIB_BIGINT_HH_
```
sgtkellox, line 26: Is this a cpplint thing or our own new standard of naming constants?

sjanssen2: I wouldn't have touched it if not cpplint is forcing me to do. It is their style.

PR #33, opened 2020-06-23
fixed some cpplint const warnings

sgtkellox:


src/parser.y
```diff
@@ -68,7 +68,7 @@ do {							\
 
 %skeleton "lalr1.cc"
 %defines  /* yacc -d */
-%define api.parser.class {Parser}
```
sjanssen2, line 71: I think you check this file in by mistake. This change is introduced by my duct tape patch solution, but should not be committed but rather reverted after building the compiler.

PR #36, opened 2020-07-07
using an updated version to generate code via bison

sjanssen2:
Currently, we are using a very dirty hack to generate code from *.y files via bison, which is necessary due to breaking changes with different bison versions. Unfortunately, different ubuntu versions or homebrew on OSX ships with those different versions. I here hope to provide a more generally working method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ jobs:
     #  dist: xenial # 16.04
     # earlier distributions are lacking libgsl-dev AND are no longer supported by Ubuntu
     - os: osx
+      osx_image: xcode11
 
 addons:
   apt:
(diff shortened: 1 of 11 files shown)


sjanssen2:
In addition to Travis test for this repo, also:
1. all builds with https://code.launchpad.net/~janssenlab/+recipe/bellmansgapc-daily were successful
2. all tests in the fold-grammars repo passed: https://travis-ci.org/github/jlab/fold-grammars/builds/705793749
3. using VirtualBox, I was able to compile gapc for OSX :-) Let's see if I can transfer that to bioconda and additional tasks in Travis

sjanssen2:
@sgtkellox and @kmaibach would be great if you could take a look at this PR. Yesterday, I got a request for gapc on OXS and thanks to the code style improvements it looks like there is actually a way to compile gapc on OSX.

sjanssen2:
This might close #38 I think the issue is that we are using an iterator to traverse the map choice_fns to simultaneously delete (all) elements from choice_fns AND the map fns. It fails because the iterator gets increased by the erase function.

PR #76, opened 2021-04-21
for window mode: ensure sub-word of input is known to algebra energy …

sjanssen2:
…functions


sjanssen2:
limiting look ahead/before for e.g. dangling bases to the current window breaks fundamental assumption of tabulation since sub-words can have different scores, depending on the left/right window border. See #79 for details

PR #78, opened 2021-04-26
test if upstream is a gap: only test left partner of closing basepair,

sjanssen2:
to avoid running out of window


sjanssen2:
The answer is: NO

PR #84, opened 2021-05-18
Add term args

sjanssen2:
extend the compiler such that the terminal parser `ROPE` can accept an argument (of type string) to restrict parses to this sequence of characters, e.g. `ROPE("kurt")` would now only yield an answer iff the sub-word is "kurt", not any other string. 

This is similar to `CHAR('A')`, but a bit more complicated since we want `ROPE` to know its now changes yield size.
Thus, `algfn(ROPE, X)` which has one moving boundary becomes `algfn(ROPE("kurt"), X)` without a moving boundary, since `ROPE` must consume exactly 4 characters.

### DIFF
--- a/src/alt.cc
+++ b/src/alt.cc
@@ -2440,7 +2440,7 @@ void Alt::Simple::init_multi_ys() {
 	assert(tracks_ == 1);
 
 	// a terminal like CHAR or ROPE can have one argument, that is an implicit filter like
-	// CHAR('a') or ROPE("kurt"), which shall only accept sub-words that are 'a' or 'kurt', respectively
+	// CHAR('a') or ROPE("stefan"), which shall only accept sub-words that are 'a' or 'stefan', respectively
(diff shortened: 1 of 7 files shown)


src/alt.cc
```diff
@@ -2437,7 +2437,28 @@ void Alt::Simple::init_multi_ys() {
   m_ys = Yield::Multi(tracks_);
 
   if (is_terminal()) {
-    assert(tracks_ == 1);
+	assert(tracks_ == 1);
+
+	// a terminal like CHAR or ROPE can have one argument, that is an implicit filter like
+	// CHAR('a') or ROPE("kurt"), which shall only accept sub-words that are 'a' or 'kurt', respectively
```
kmaibach, line 2443: you chose ROPE("stefan") in the other files as an example - I'd change the "kurt" ones into "stefan" for the sake of consistency

kmaibach (approved):

PR #96, opened 2022-01-21
only execute checks in debug mode

sjanssen2:
This is to fix warnings when **not** compiling in debug mode (assertions are not used):
```
g++  -I/usr/include   -O3  -std=c++17 -D_XOPEN_SOURCE=500 -MMD -MP -Wall -Wnon-virtual-dtor -Wno-unused-variable -Wno-parentheses -DNDEBUG -DBISONNEW   src/backtrack.cc -c -o src/backtrack.o 
src/backtrack.cc: In member function ‘virtual void Backtrack::gen_backtrack(AST&)’:
src/backtrack.cc:77:8: warning: variable ‘r’ set but not used [-Wunused-but-set-variable]
   bool r = ast.check_instances(instance);
        ^
```

### DIFF
--- a/src/backtrack.cc
+++ b/src/backtrack.cc
@@ -74,9 +74,11 @@ void Backtrack::apply_filter(Filter *f) {
 
 
 void Backtrack::gen_backtrack(AST &ast) {
+  #ifndef NDEBUG
   bool r = ast.check_instances(instance);
   assert(r);
   r = ast.insert_instance(instance);
+  #endif
   assert(r);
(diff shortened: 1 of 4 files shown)


sjanssen2:
see deatils here: https://en.cppreference.com/w/cpp/language/attributes/maybe_unused
and usage here: https://stackoverflow.com/questions/777261/avoiding-unused-variables-warnings-when-using-assert-in-a-release-build

PR #97, opened 2022-01-21
avoid deprecated auto_ptr

sjanssen2:


sjanssen2:
see https://stackoverflow.com/questions/45053626/error-templateclass-class-stdauto-ptr-is-deprecated

PR #108, opened 2022-01-30
flag one parameter constructors explicit

sjanssen2:
Reduce number of runtime/explicit cpplint warnings. Starting at 160 for all files in src/*


sjanssen2:
Category 'runtime/explicit' errors found: 14

PR #113, opened 2022-02-03
Rnahybrid 2021

sjanssen2:
In order to reproduce the original MFE computation of RNAhybrid (Haskell and ADPc versions), we need to expose some additional energy contribution functions in the librna.


sjanssen2:
Hi @kmaibach I think it is a good time to merge this PR, i.e. the exposure of those energy functions necessary for RNAhybrid, to ease your development. Thus, you can use the "default", i.e. master branch version of gapc.

kmaibach (approved):

PR #119, opened 2022-03-17
Plot Grammar

sjanssen2:
This PR adds a novel feature. With `--plot-grammar`, the user can request the generation of a out.dot file that holds a rudimentary graph representing the grammar. This is handy if the grammar is modified by the compiler.

This is not yet complete! Missing:
  - [x] handling of multi-track
  - [x] annotation of semantic filtering
  - [x] unit tests
  - [x] index hacks, aka implicit indices: 8-sided polygons for nodes + red bold explicit indices
  - [x] tabulation: non-terminals NOT resulting in DP-tables are plotted in dotted rectangles
  - [x] multi-filter
  - [x] evaluation function

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -2,9 +2,9 @@ name: github action build & CI
 
 on:
   push:
-    branches: [ master, plot_grammar, plot_grammar_style ]
+    branches: [ master, plot_grammar, plot_grammar_polish ]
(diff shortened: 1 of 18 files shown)


sjanssen2:
@Jasmin-Walter we talked about indicating if a non-terminal is tabulated or not. I've added the according code to draw these rectangles with "dotted" lines

PR #123, opened 2022-04-08
spare else statement with empty(ret_x) which will get replaced to emp…

sjanssen2:
…ty(answers) and thus delete previously computed valid candidates

### DIFF
--- a/src/alt.cc
+++ b/src/alt.cc
@@ -2264,8 +2264,15 @@ void Alt::Base::init_filter_guards(AST &ast) {
   Expr::Base *arg = Expr::seq_to_tree<Expr::Base, Expr::And>(
     exprs.begin(), exprs.end());
   Statement::If *guard = new Statement::If(arg);
-  guard->els.push_back (new Statement::Fn_Call(
-    Statement::Fn_Call::EMPTY, *ret_decl));
+  if ((this->data_type()->is(::Type::LIST)) && (this->top_level)) {
+    // don't add an else statement (which erases the return type) to the
+    // filter since this would "empty"=erase the answer list of ALL
+    // alternatives of a symbol due to the return decl replacement happening
+    // in Symbol::NT::eliminate_list_ass
+  } else {
+    guard->els.push_back(new Statement::Fn_Call(
+      Statement::Fn_Call::EMPTY, *ret_decl));
+  }
   filter_guards = guard;
 }
 


sjanssen2:
looks like this issue persists also in other locations, namely the application of filters on Alt::Links (i.e. alternatives that directly point to other non-terminals without using algebra functions). Again, we need to ensure that the called parser returns not a single element but a list, i.e. have at least one moving boundary. Here is a minimal example:

```
import rna

input rna

signature sig_foldrna(alphabet,answer) {
	answer nil(Subsequence); //empty structure
	answer bl(Subsequence, Subsequence, Subsequence, Subsequence); // a bulge loop to the left with a closing base-pair
	choice [answer] h([answer]);
}

algebra alg_count auto count;
algebra alg_enum auto enum;

algebra alg_mfe implements sig_foldrna(alphabet = char, answer = int) {
  int bl(Subsequence lb, Subsequence lr, Subsequence x, Subsequence rb) {
    return 2;
  }
  int nil(Subsequence n) {
    return 1;
  }
  choice [int] h([int] i) {
    return list(minimum(i));
  }
}

grammar gra_nodangle uses sig_foldrna(axiom = struct) {
  struct    = leftB |
              leftB with minsize(5)   |
              nil(LOC) # h;

  leftB     = bl(BASE, REGION, REGION, BASE) with basepairing # h;
}

instance ins = gra_nodangle(alg_mfe * alg_enum);
```
Executing the resulting compiled program (instance `ins`) with input `CaaG` leads to an error:
`out: ../rtlib/push_back.hh:196: void append_min_other(List_Ref<T, pos_int>&, List_Ref<T, pos_int>&) [with T = std::pair<int, rope::Ref<rope::Ref_Count> >; pos_int = unsigned char]: Assertion `&x.ref() != &e.ref()' failed.`

The issue is located here:
```
  List_Ref<std::pair<int, Rope> > ret_1 = answers;
  if (minsize(t_0_seq, t_0_left_most, t_0_right_most, 5)) {
    ret_1 = nt_leftB();
  }
  append_min_other(answers, ret_1);
```
Why should `ret_1` set to `answers`? If the filter fails, we would append `answers` to `answers`. I think the `= answers` assignment should be avoided AND we should guard the `append_min_other` with an `if (is_not_empty(ret_1))`.

PR #128, opened 2022-05-12
fix TODO and use Log mechanism to report table-design and runtime

sjanssen2:


kmaibach (approved):
Which log level do I have to select to get this message printed? 
And could we include that information in the help page (or any other documentation place) somehow?

PR #142, opened 2022-09-23
Backward

sjanssen2:
This is a quick and dirty attempt to compute 1st and 2nd derivatives.

- increase table capacity to additionally store `q`'s `e`'s from backward pass
  - will this work in window mode?
  - what about non tabulated NTs?
  - check in Symbol::NT::set_ret_decl_rhs if other cases also need a "track" call
  - can we have situations in which same q edge is used multiple times but with different weights? `ntX = add(ntA, ntB, ntC)` such that moving bundary between `ntB and btC` with same `ntA` is used?


sjanssen2:
outdated by #151

PR #153, opened 2022-11-01
Speed up calculations in energy functions

fymue:
The `getNext` and `getPrev` functions, which are implemented in the [rnalib.c](https://github.com/jlab/gapc/blob/master/librna/rnalib.c) source file and are used to calculate the index/sequence position of the next non-gap base in a given range/region of the input sequence/alignment, are usually called a lot during the runtime of Gapc compilations. Most of those calls are with identical function parameters. Hoping that we might be able to save some execution time in the Gapc compilations, I tried to reuse the calculation results of these functions by storing them in a lookup array and thus to avoid multiple recalculations of the same results. Unfortunately, it seems as if the index calculation required to retrieve the correct lookup value takes about as long as the calculations the functions actually carry out, which negates any potential performance gain and actually slightly increases the total execution time of the program I tested (blue line in plot below).

Example call used during benchmark (to increase alignment size, I concatenated each alignment row to itself a few times): 
`./out GAGGGUGCCUCUGACCGUGCCCG_____AUGAC___ACCAACCUGCCCAAUGAAGCACGGGCGACCGUGGGGA___GGGUGUU_______________________________________CCCUGUGGGAACCCUCGCCAUUCACGGCGAGGAGGAGGUCAGCUCCUCCACGUUC#GAGGGCUG_UCUGACCCAGCCCG_____AUGAG___CUGAACCUGCCCUGUGACCUACCGGCGACGGUAGGGAGAAGGGUGUU_______________________________________CCCCGUGGGAACCUCCGAGCUUUAGCUCGGAGAGGAGGUCAGCUUUUAACAUUUC#GAGGGCUG_UCUGACCCAGCCCG_____AUGAG___CUGAACCUGCCCUGUGACCUACCGGCGACGGUAGGGAGAAGGGUGUU_______________________________________CCCCGUGGGAACCUCCGAGCUUUAGCUCGGAGAGGAGGUCAGCUUUUAACAUUUC#GUGGUCUUUAC_GCCCAAGGGAGGGGUAAUGGGCCUGAGACCGGGCCC_UCGGGC______CGAAC___CGAAACC_GGUG__ACGGGAGUAGGUGGAUGAGAGCCCGCAAACCUCUCCCCGCCC_GUUGAAAGCCUCGCCCUUCAAGGCGGGGAGGAGGUCAGGAGG_AGAUAACC#GAGGGUAGCUCUGACUCUGCCCG_____AUGAC___ACCAACCUGCCCUGUGACCUAUCGGCGACGAUAAGUAGCAGGGUGUU_______________________________________CCGUGUUGGAACCUCCACCCUUUAGGU_GGAGAGAAGAUCAG___UUAAAAUUCC#`

mfe_benchmark_time

Interestingly, the overall impact of the `getNext` and `getPrev` functions on the total execution time doesn't seem to be big at all, at least when using the example alignment that I used (see example call). As you can see in the plot above, the execution time of the program when modifying `getNext` and `getPrev` by writing the statement `return 1;` as the first line in both functions (red line), which essentially makes them as fast as they could possibly be (and obviously wrong; but this is only an example), barely improves the execution time over the currently used implementation (green line).

To summarize: Even though it made sense to assume that there is/was optimization potential in regards to reusing `getNext` and `getPrev` calculation results, the actual calculations carried out in these functions seem to require too little time to make the additional overhead of lookup index calculation etc. worthwhile, which is actually a good thing though, because that means that the functions are already about as fast as they could possibly be.

### DIFF
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -429,12 +429,12 @@ double mk_pf(double energy) {
 
 static double get_scale_value(int subword_size) {
   // precalculate the first 10000 scale values
-  
+
   static bool init = false;
   static const double MEAN_NRG = -0.1843;
   static double lookup[LOOKUP_SIZE];
   static double mean_scale;
-  
+
   if (!init) {
     // temperature can be adjusted with the -T flag at runtime
     mean_scale = exp(-1.0 * MEAN_NRG /
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -501,7 +501,7 @@ static rsize getPrev(const char *s, rsize pos, rsize steps, rsize leftBorder) {
 
   if ((pos <= leftBorder) || (x <= leftBorder))
     return leftBorder;
-  
+
   rsize nongaps = 0;
 
   do {
@@ -1157,16 +1157,18 @@ int ml_mismatch_energy(const char *s, rsize i, rsize j) {
 static rsize getNext(const char *s, rsize pos, rsize steps,
                      rsize rightBorder, unsigned currRow, unsigned maxRows) {
   assert(steps > 0);
-  rsize nongaps = 0;
+
   rsize x = pos + 1;
   if (x >= rightBorder)
     return rightBorder;
+
+  rsize nongaps = 0;
 
   static bool init = false;
   static rsize *lookup;
   static unsigned seqLen;
   static unsigned triu;
-  
+
   if (!init) {
     seqLen = strlen(s);
     triu = (seqLen * (seqLen + 1)) / 2 + 1;
@@ -1176,8 +1178,9 @@ static rsize getNext(const char *s, rsize pos, rsize steps,
   }
 
   unsigned long index = (steps - 1) * triu * maxRows + triu * currRow +
-                        pos * seqLen + rightBorder - ((pos * (pos + 1)) / 2);
-  
+                        pos * seqLen + rightBorder -
+                        ((pos * (pos + 1)) / 2);
+
   if (lookup[index]) {
     return lookup[index];
   }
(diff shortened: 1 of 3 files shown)


sjanssen2:
My own evaluation (outside alg_pfunc) confirms the little to no gain of your current implementation :-/
image
image

fymue:
@sjanssen2  Since you also wanted me to check whether or not it makes sense to precalculate/lookup calculated results of some of the energy functions (`il_energy`, `br_energy`, `bl_energy` and `hl_energy`) of the [rnalib.c](https://github.com/jlab/gapc/blob/master/librna/rnalib.c)-file , I implemented multiple versions of a hashmap-based cache for those functions (an array-based cache/lookup unfortunately isn't possible here, because these energy functions have too many variable parameters which would make arrays that can fit all possible results way too large), benchmarked those and got the following results:

energy_benchmark_space

energy_benchmark_time

As you can see, caching previously calculated results and reusing them later actually increases the execution time by a fairly large amount compared to the current/old implementation (blue line) while also majorly increasing the required memory. This slowdown when reusing previously calculated results actually makes sense though when looking at the execution times of the maximum possible speedup for these energy functions (red line), which are only slightly lower than the execution times of the current implementation of these functions (blue line).  So at least for this Gapc compilation (`gapc -p alg_pfunc nodangle.gap`), the calculations of the energy functions don't seem to have a major impact on the overall execution time. And even if their impact is bigger in other Gapc compilations, I doubt that these particular functions are a good candidate for caching/lookups, because the hashmap-based cache most likely creates too much computational overhead while also requiring way too much memory and an array-based cache is probably not possible here due to the many variable parameters of those functions.

PR #154, opened 2022-11-02
precalculate the first 10000 scale values

fymue:
In order to reduce the execution time of some Gapc compilations, we figured it would make sense to precalculate (most of) the required return values of the `scale` function from the [rnalib.c](https://github.com/jlab/gapc/blob/master/librna/rnalib.c)-file, since this function is called quite often in the return statements of certain algebra functions. As expected, precalculating the scale values reduces the execution time of the Gapc compilation I tested (`gapc -p alg_pfunc nodangle.gap`) by a significant amount (see benchmark below).

scale_benchmark

I once again tested multiple scale value lookup implementations (dynamic lookup (blue line), full precalculation) and concluded that precalculating the scale values beforehand as well as moving this process to a dedicated function reduces the execution time the most (red line). It is a little bit suprising to see that moving the precalculation to a dedicated function improves the execution speed over doing this process directly in the `scale` function (green line) though. There probably must be some sort of compiler optimization happening there.

### DIFF
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -55,6 +55,8 @@
 #include "ViennaRNA/params/io.h"
 #include "ViennaRNA/fold_vars.h"
 
+#define LOOKUP_SIZE 10000
+
 static vrna_param_t  *P = 0;
 
 
@@ -1076,28 +1078,28 @@ double mk_pf(double x) {
 
 double getScaleValue(int x) {
   static bool init = false;
-  static const double mean_nrg = -0.1843;
-  static double lookup[10000];
+  static const double MEAN_NRG = -0.1843;
+  static double lookup[LOOKUP_SIZE];
   static double mean_scale;
 
   if (!init) {
     // precalculate the first 10000 scale values
-    mean_scale = exp(-1.0 * mean_nrg /
+    mean_scale = exp(-1.0 * MEAN_NRG /
                      (GASCONST / 1000 *
                       (temperature + K0)));
 
-    for (int i = 0; i < 10000; i++) {
+    for (int i = 0; i < LOOKUP_SIZE; i++) {
       lookup[i] = 1.0 / pow(mean_scale, i);
     }
 
     init = true;
   }
 
-  if (x < 10000) {
+  if (x < LOOKUP_SIZE) {
     return lookup[x];
   } else {
     /* in the rare cases that the required scale value
-       is bigger than or equal to 10000, calculate the
+       is bigger than or equal to LOOKUP_SIZE, calculate the
        value on the spot */
     return 1.0 / pow(mean_scale, x);
   }
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -45,6 +45,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <math.h>
+#include <stdbool.h>
 
 #include "ViennaRNA/datastructures/basic.h"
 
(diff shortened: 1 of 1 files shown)


librna/rnalib.c
```diff
@@ -1073,16 +1074,41 @@ double mk_pf(double x) {
   return exp((-1.0 * x/100.0) / (GASCONST/1000 * (temperature + K0)));
 }
 
+double getScaleValue(int x) {
+  static bool init = false;
+  static const double mean_nrg = -0.1843;
+  static double lookup[10000];
```
sjanssen2, line 1080: can we please define a const variable for the threshold of 10000, since it has to be in sync with line 1096 and 1089

librna/rnalib.c
```diff
@@ -1073,16 +1074,41 @@ double mk_pf(double x) {
   return exp((-1.0 * x/100.0) / (GASCONST/1000 * (temperature + K0)));
 }
 
+double getScaleValue(int x) {
+  static bool init = false;
+  static const double mean_nrg = -0.1843;
```
sjanssen2, line 1079: can we refactor the name to ALL CAPS since it is a constant?

sjanssen2 (changes requested):

librna/rnalib.c
```diff
@@ -45,6 +45,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <math.h>
+#include <stdbool.h>
```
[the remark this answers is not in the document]
fymue: This include statement is needed in order to be able to use  the `bool` datatype in `getScaleValue`. `bool` doesn't exist by default in C.

librna/rnalib.c
fymue: Unfortunately C doesn't allow the initialization of static arrays with a non-constant size. It doesn't even regard a `const` variable as that and throws a compiler error when using a `const` variable as the array storage size. What I could do is allocate the array on the heap using `malloc` and simpy assign the resulting pointer to a static pointer, which will then serve as the array, like so:
```
static double* lookup;
static const int size = 10000;
if (!init) {
  lookup = malloc(size * sizeof(double);
  ...
}
```

This would work, but obviously means that we would have to allocate the array on the heap, which takes a little longer than a stack allocation. But this is probably negligible.

librna/rnalib.c
fymue: Sure, will do.

librna/rnalib.c
fymue: A macro would work as well.

sjanssen2 (approved):

sjanssen2:
again, two of my benchmarks (ignore blue, orange is master, green is this PR):
image
image
in conclusion: we treat a tiny fraction of memory for a slight speed up. But every second counts ;-)

librna/rnalib.c
```diff
@@ -54,6 +55,8 @@
 #include "ViennaRNA/params/io.h"
 #include "ViennaRNA/fold_vars.h"
 
+#define LOOKUP_SIZE 10000
```
kmaibach, line 58: Why 10k? Is there a reason for that number?

kmaibach (approved):

librna/rnalib.c
fymue: The biggest possible number that can be passed to the `scale` function during runtime is the length of the input sequence. Unfortunately passing the length of the input sequence to `scale` would've been too complicated, so we assume that input sequences should be shorter than 10000 bases most of the time, which is why we chose to precalculate the first 10000 scale values. If the input sequence is longer than 10000 bases, the scale values >= 10000 will be calculated on the spot.

PR #156, opened 2022-11-07
precalculate mk_pf energy scaling values

fymue:
Similarly to the last PR #154 , I figured it would also makes sense to precalculate a relatively large amount of the scaled energy values the `mk_pf` function from the [rnalib.c](https://github.com/jlab/gapc/blob/master/librna/rnalib.c)-file returns, since this function is also called quite often in the return statements of certain algebra functions.

mkpf_benchmark

As you can see in the benchmark plot above, precalulating just the first 10000 `mk_pf` scaled energy values reduces the execution time of the Gapc compilation I tested (`gapc -p alg_pfunc nodangle.gap`) by about as much as is theoretically possible for this particular function (see green line for maximum possible speedup), while only minimally increasing the required memory (by such a small amount that you can't even see it in the memory usage plot below).

mkpf_benchmark_space

### DIFF
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -1067,13 +1067,48 @@ int ss_energy(rsize i, rsize j) {
   return 0;
 }
 
+double getMkpfValue(int x) {
+  /* precalculate LOOKUP_SIZE mk_pf values;
+     since input values can be negative,
+     values within the range (-HALF, HALF)
+     will be calculated (HALF = LOOKUP_SIZE / 2) */
+
+  static bool init = false;
+  static const int HALF = LOOKUP_SIZE / 2;
+  static double lookup[LOOKUP_SIZE];
+  static double divisor;
+
+  if (!init) {
+    // temperature is defined in ViennaRNA/params/basic.h
+    divisor = GASCONST / 1000 * (temperature + K0);
+
+    // calculate mk_pf values from -HALF to HALF
+    for (int i = -HALF; i < HALF; i++) {
+      lookup[i + HALF] = exp((-1.0 * i / 100.0) / divisor);
+    }
+    
+    init = true;
+  }
+
+  int index = x + HALF;
+
+  if (index > 0 && index < LOOKUP_SIZE) {
+    return lookup[index];
+  } else {
+    return exp((-1.0 * x / 100.0) / divisor);
+  }
+}
 
 /*
    scales the energy value x into a partition function value
 */
 double mk_pf(double x) {
-  // temperature is defined in ViennaRNA/params/basic.h
-  return exp((-1.0 * x/100.0) / (GASCONST/1000 * (temperature + K0)));
+  /* getMkpfValue takes an argument of type int instead of
+     type double, because the energy functions
+     (e.g. il_energy) actually return int instead of double;
+     so we let C do an implicit conversion here */
+
+  return getMkpfValue(x);
 }
 
 double getScaleValue(int x) {
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -1090,6 +1090,8 @@ static double get_mkpf_value(double x) {
     init = true;
   }
 
+  return exp((-1.0 * x / 100.0) / divisor);
+
   int index = x + HALF;
 
   if (index >= 0 && index < LOOKUP_SIZE) {
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -1090,11 +1090,9 @@ static double get_mkpf_value(double x) {
     init = true;
   }
 
-  return exp((-1.0 * x / 100.0) / divisor);
-
   int index = x + HALF;
 
-  if (index >= 0 && index < LOOKUP_SIZE) {
+  if (index >= 0 && index < LOOKUP_SIZE && trunc(x) == x) {
     return lookup[index];
   } else {
     /* if the input energy value isn't within the range
--- a/librna/rnalib.c
+++ b/librna/rnalib.c
@@ -1067,7 +1067,7 @@ int ss_energy(rsize i, rsize j) {
   return 0;
 }
 
-static double get_mkpf_value(double x) {
+static double get_mkpf_value(double energy) {
   /* -precalculate LOOKUP_SIZE mk_pf partition function values
      -since input values can be negative,
       values within the range -HALF <= x < HALF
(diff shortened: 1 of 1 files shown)


librna/rnalib.c
```diff
@@ -1067,16 +1067,49 @@ int ss_energy(rsize i, rsize j) {
...
 
+double get_mkpf_value(int x) {
+  /* -precalculate LOOKUP_SIZE mk_pf partition function values
+     -since input values can be negative,
+      values within the range -HALF <= x < HALF
+      will be calculated (HALF = LOOKUP_SIZE / 2) */
+
+  static bool init = false;
+  static const int HALF = LOOKUP_SIZE / 2;
+  static double lookup[LOOKUP_SIZE];
```
kmaibach, line 1078: Regarding the LOOKUP_SIZE (which is 10,000 right?): Is it a given that the array `lookup` has always at least 10k entries? Or could it be possible that it's smaller - leading to an "index out of bounds" error?

kmaibach: Ah, never mind my question, I misunderstood the code.

librna/rnalib.c
```diff
@@ -1090,11 +1090,9 @@ static double get_mkpf_value(double x) {
     init = true;
   }
 
-  return exp((-1.0 * x / 100.0) / divisor);
-
   int index = x + HALF;
 
-  if (index >= 0 && index < LOOKUP_SIZE) {
+  if (index >= 0 && index < LOOKUP_SIZE && trunc(x) == x) {
```
fymue, line 1095: Adding this check (whether the `double x` input parameter has a remainder equal to 0) fixed the bug/problem. I assumed that the `mk_pf` function was only ever called to scale the `int` values coming from the energy functions (`il_energy` etc.). But apparently there are also some `mk_pf` calls with actual `double` values and not just with `int` values converted to `double` values, which is why the function was performing lookups for some input values it didn't actually precalculate.

kmaibach:
Do you also see that 2 checks from github actions are not completed although they are marked as completed if you look at the checks in the `Actions` tab?
Is this some kind of bug or normal behaviour?
Not really important, just something that caught my eye.

fymue:
Yeah, these two checks didn't start for the commits of the last PR #154 either, not sure why. But apparently @sjanssen2 didn't care about that since PR #154 has already been merged.

sjanssen2:
These two checks are runs on deprecated OS versions. Github no longer supports ubuntu-18.04 or macos-10.15. However, I wonder why these runs show up in the list of checks ?!? But yes, we don't have to care!

sjanssen2 (changes requested):
hm, I am not sure if I understand this approach? Input for get_mkpf_value is e.g. `mk_pf(termau_energy(lb, rb))` from the algebra pfunc, i.e. free energy values. These stem ultimately from the NNDB tables, e.g. http://rna.urmc.rochester.edu/NNDB/turner04/stack.txt which means the input values should NOT be a continuum. But you iterate continuously from -5000 to 5000. How many of these values are actually requested? I would assume that this might be quite sparse. Have you experimented with a Hashmap instead of an array?

librna/rnalib.c
fymue: Unfortunately I'm not familiar with these rules so I can't answer to that, sorry. I didn't encounter this problem during any of my local tests, but one Action kept failing and i eventually figured out that this was the reason. If this function is only ever called with `int` values then this is really confusing since all energy values should indeed fit into `int` values, so i don't exactly know in which situations this check actually would return `false`.

librna/rnalib.c
```diff
@@ -1067,16 +1067,49 @@ int ss_energy(rsize i, rsize j) {
   return 0;
 }
 
+static double get_mkpf_value(double x) {
```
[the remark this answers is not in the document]
fymue: Sure, I will update the variable names.

kmaibach:
> These two checks are runs on deprecated OS versions. Github no longer supports ubuntu-18.04 or macos-10.15. However, I wonder why these runs show up in the list of checks ?!? But yes, we don't have to care!

Could we maybe delete or exclude those checks if they are not working anymore anyways?

librna/rnalib.c
```diff
@@ -1067,16 +1067,49 @@ int ss_energy(rsize i, rsize j) {
...
+
+  int index = x + HALF;
+
+  if (index >= 0 && index < LOOKUP_SIZE && trunc(x) == x) {
+    return lookup[index];
+  } else {
+    /* if the input energy value isn't within the range
+       -HALF <= x < HALF (meaning the index isn't within the range
+       0 <= i < LOOKUP_SIZE), calculate the mk_pf value on the spot */
+    return exp((-1.0 * x / 100.0) / divisor);
```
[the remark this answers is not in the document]
fymue: The first time this function is called `init` is guaranteed to be `false`, so `divisor` will be defined then. There is no scenario I can think of in which the `return` statement can be reached before `divisor` is defined.

librna/rnalib.c
```diff
@@ -1110,7 +1143,7 @@ double getScaleValue(int x) {
 */
 double scale(int x) {
```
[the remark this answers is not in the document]
fymue: That would makes sense, will update.

librna/rnalib.c
```diff
@@ -1067,16 +1067,49 @@ int ss_energy(rsize i, rsize j) {
...
+  static const int HALF = LOOKUP_SIZE / 2;
+  static double lookup[LOOKUP_SIZE];
+  static double divisor;
+
+  if (!init) {
+    // temperature is defined in ViennaRNA/params/basic.h
+    divisor = GASCONST / 1000 * (temperature + K0);
+
+    // calculate mk_pf partition function values in range -HALF <= x < HALF
+    for (int i = 0; i < LOOKUP_SIZE; i++) {
```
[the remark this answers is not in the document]
fymue: Sure, we can let `i` start from `-HALF` and iterate as long as `i < HALF`. I will adjust that.

fymue:
> hm, I am not sure if I understand this approach? Input for get_mkpf_value is e.g. `mk_pf(termau_energy(lb, rb))` from the algebra pfunc, i.e. free energy values. These stem ultimately from the NNDB tables, e.g. http://rna.urmc.rochester.edu/NNDB/turner04/stack.txt which means the input values should NOT be a continuum. But you iterate continuously from -5000 to 5000. How many of these values are actually requested? I would assume that this might be quite sparse. Have you experimented with a Hashmap instead of an array?

I agree, a lot of the calculated values are probably never used. I only noticed that both positive and negative energy values are passed to this function, so I figured I would just precalculate as many possible values as I could within the `LOOKUP_SIZE` range, so I simply split it in half. If the input energy values repeat a lot, meaning only a few distinct values are used, we could also dynamically store those, so only values that are required at least once will be added to the lookup array. But since I only calculate 10000 values here (which is essentially instantaneous) there probably would be little to no performance improvement with this new approach. But I will try it out just in case.
Switching from array to a hashmap is probably not worth looking into though, since hashing the input energy value and performing the lookup would most likely take longer than just calculating the required value on the spot.

librna/rnalib.c
sjanssen2: might point to a general bug, thus it could be worth following it. Can you open and describe this problem as a new issue?

fymue: I opened an issue addressing this problem.

fymue:
I just got done benchmarking an adapted `get_mkpf_value` implementation which only stores/performs lookups on energy values that were passed to it in a previous function call (and doesn't precalculate all values in the range [-5000, 5000) beforehand):

mkpf_benchmark_precalc_vs_dynamic

As you can see, both implementations are about equally fast (blue line is from first benchmark shown in this PR). I also measured the time it takes to precalculate the 10000 scaled energy values, and that takes only about 260 μs (0.000260s). So I think we could just keep the new implementation as it is, even if we precalculate a lot of unused values.

librna/rnalib.c
```diff
@@ -1067,23 +1067,60 @@ int ss_energy(rsize i, rsize j) {
...
+  }
+}
 
 /*
    scales the energy value x into a partition function value
 */
-double mk_pf(double x) {
-  // temperature is defined in ViennaRNA/params/basic.h
-  return exp((-1.0 * x/100.0) / (GASCONST/1000 * (temperature + K0)));
+double mk_pf(double energy) {
```
sjanssen2, line 1110: why does `mk_pf` only call `get_mkpf_value`? Can't we simply rename your `get_mkpf_value` function into `mk_pf` and thus replace my old implementation?

sjanssen2 (changes requested):

sjanssen2:
> So I think we could just keep the new implementation as it is, even if we precalculate a lot of unused values.

agreed. Thanks for testing!

librna/rnalib.c
fymue: I implemented it this way analogous to the precalculation of the `scale` function values (see PR #154), where for some reason encapsulating the code in another function and calling that function in the actual `scale` function was noticeably faster.
I could of course rename `get_mkpf_value` to `m̀k_pf` and skip the encapsulation part, but I for one assume that this weird slight speedup we saw in PR #154 when encapsulating would also happen here and also just wanted to continue that style. If you want to do all of the work in just one function we can of course do that, but in that case I would say we also do the same for `get_scale_value` and `scale`.

sjanssen2: oh, forgot about that curiosity. Sorry. So let's keep as is!

sjanssen2 (approved):

PR #158, opened 2022-11-26
ViennaRNA v2.5.1 update

fymue:
# PR Description

This PR contains the latest updates of the ViennaRNA source files we use in Gapc. I appended a changelog with a list of the major changes the newest ViennaRNA release (v2.5.1) introduced over the currently used release in Gapc (v2.4.17).

# Changelog Gapc ViennaRNA v2.5.1 update

The following ViennaRNA files have been modified since the currently used ViennaRNA release in Gapc (v2.4.17, 11/2020):

- gapc/librna/ViennaRNA/
    - `alphabet.c`: changed some things in the `vrna_ptypes_prepare` function
    - `model.h`: added `pair_dist` matrix declaration
    - `model.c`: refactored some constant matrices to macros
- gapc/librna/ViennaRNA/constraints/
    - `hard.h`: removed `matrix` variable
    - `soft.h`: many `void` functions now return `int`
- gapc/librna/ViennaRNA/utils/
    - `basic.h`: added signature of `vrna_init_rand_seed` function
    - `string_utils.c`: added new functions `vrna_seq_reverse`, `vrna_DNA_complement`, `vrna_strtrim`
    - `strings.h`: added headers and macros for the new functions in `string_utils.c`
    - `structures.h`: renamed some parameters, added some function signatures
    - `utils.c`: added the `vrna_init_rand_seed` implementation
- gapc/librna/ViennaRNA/static/misc/
    - updated `dna_mathews1999.hex`, `rna_andronescu2007.hex`, `rna_langdon2018.hex` and `rna_turner1999.hex` since the analogous `.par` files have been adjusted
- gapc/librna/paramfiles/
    - replaced DEF placeholders in parameter files with their actual value -50 in `dna_mathews1999.par`, `rna_andronescu2007.par`, `rna_langdon2018.par` and `rna_turner1999.par`


fymue:
> There are a few things in the updated ViennaRNA code that `cpplint` regards as errors which aren't really codestyle-related errors though but rather the way the ViennaRNA people wrote their code (which `cpplint` doesn't seem to like all too much). If I wanted to fix of all these errors I would have to change the datatypes of some variables and ways they implemented some of their code, and I don't know if that's such a good idea. The list below contains all additional filters I would recommend adding to the `cpplint` action along with a small description of what it complains about:
> 
>     * `-runtime/int`: wants you to use `int64_t` instead of `long`, `int16_t` instead of `short` etc.
> 
>     * `-legal/copyright`: Copyright message missing
> 
>     * `-whitespace/empty_loop_body`: wrongfully thinks `do-while` loops are `while` loops with empty body; seems to be an actual [bug](https://github.com/ros/roslint/issues/28)
> 
>     * `-runtime/printf`: wants you to use `snprintf` for all string-copy-related things (for safety reasons I assume)
> 
> 
> I can of course try to adjust the ViennaRNA code so `cpplint` is happy, but assuming some of the fold-grammars tests fail due to the code changes of the new ViennaRNA release, we then don't know if it was caused by an error on my part or simply because the new ViennaRNA code is still incompatible with the current test cases. Your call @sjanssen2.
 
I just noticed that the `cpplint` action apparently doesn't even lint these files, so ignore the `cpplint` part.

sjanssen2:
For exactly the reasons you mentioned ;-) Since it is not our code  I think
it's fine to leave code styling to the Vienna guys

Fynn Mürmanns ***@***.***> schrieb am Sa., 26. Nov. 2022,
20:17:

fymue:
Ok, makes sense. All but the first commit I pushed to this branch were codestyle "fixes". So if we let ViennaRNA guys use their codestyle, should I undo those changes and go back to the initial commit before we merge?

sjanssen2:
> Ok, makes sense. All but the first commit I pushed to this branch were codestyle "fixes". So if we let ViennaRNA guys use their codestyle, should I undo those changes and go back to the initial commit before we merge?

Hi @fymue, I'd vote for reverting. Thus, we can keep the external files from the Vienna source base as is. Thanks!

sjanssen2 (changes requested):
Looks great! And I am surprised that no energies changed, i.e. all tests pass :-)
Could you please leave a comment about the update, here:
- https://github.com/jlab/gapc/blob/master/librna/readme
- https://github.com/jlab/gapc/blob/e04b5ca03256dfa8559a9fa8cff9a791f6d9192d/librna/rnalib.c#L31
And double check if we have to update the configuration during installation? We should at least change the version numbers
- [config.h](https://github.com/jlab/gapc/blob/e04b5ca03256dfa8559a9fa8cff9a791f6d9192d/librna/configure.ac)

fymue:
I added a comment about the update in `rnalib.c` . I also updated the version number in the `configure.ac` file and auto-generated an updated `configure` file. Everything compiled on my end, so I don't think we have to adjust anything else regarding the configuration.

fymue:
> Looks great! And I am surprised that no energies changed, i.e. all tests pass :-)

 @sjanssen2 I just noticed that some energy parameter files in the `librna/paramfiles` directory actually have been modified as well (not in v2.5.1, but in v.2.4.18). I guess I must have missed those files when scanning for changes, because they are in a different directory in the ViennaRNA repo. I will update them ASAP.

sjanssen2 (approved):

sjanssen2 (changes requested):
I did a diff on the source folders of gapc and ViennaRNA and found diverging files:
- ViennaRNA-2.5.1/src/ViennaRNA/static/misc/dna_mathews1999.hex
- ViennaRNA-2.5.1/src/ViennaRNA/static/misc/rna_andronescu2007.hex
- ViennaRNA-2.5.1/src/ViennaRNA/static/misc/rna_langdon2018.hex
- ViennaRNA-2.5.1/src/ViennaRNA/static/misc/rna_turner1999.hex
-

fymue:
> I did a diff on the source folders of gapc and ViennaRNA and found diverging files:
> 
>     * ViennaRNA-2.5.1/src/ViennaRNA/static/misc/dna_mathews1999.hex
> 
>     * ViennaRNA-2.5.1/src/ViennaRNA/static/misc/rna_andronescu2007.hex
> 
>     * ViennaRNA-2.5.1/src/ViennaRNA/static/misc/rna_langdon2018.hex
> 
>     * ViennaRNA-2.5.1/src/ViennaRNA/static/misc/rna_turner1999.hex
> 
> 
> Could you please update those as well?

I tried to figure where these `.hex`-files are coming from, but I couldn't seem to find their source. Since they are nowhere in the ViennaRNA repo (I assume you build ViennaRNA before the diff?), I scanned the ViennaRNA Makefiles and found this [Makefile](https://github.com/ViennaRNA/ViennaRNA/blob/master/src/ViennaRNA/static/Makefile.am), in which Line 32 tells me that the `.hex`-files are automatically generated from the (updated) `.par` files:
```
.par.hex:
	$(xxd_verbose)$(MKDIR_P) misc && $(XXD) -i < $< > $@ && echo ',0x00' >> $@
```
This doesn't happen when I rebuild `gapc`, which probably makes sense though since we don't have a `Makefile` specifically for the ViennaRNA subdir (I think?). Copying the [Makefile](https://github.com/ViennaRNA/ViennaRNA/blob/master/src/ViennaRNA/static/Makefile.am) command, I generated the updated `.hex`-files manually, e.g. like:
```
xdd -i librna/paramfiles/dna_mathews1999.par > librna/ViennaRNA/static/misc/dna_mathews1999.hex
echo ',0x00' >> librna/ViennaRNA/static/misc/rna_turner1999.hex
```
I then had to remove some of the generated C header stuff like the variable name etc., since we only store the array content in the `.hex`-files and directly include the content in `ViennaRNA/static/energy_parameter_sets.h`. I hope this approach was correct, please tell me if I did something wrong here.

sjanssen2:
The `configure` file in `librna` should be called from the `configure` script in the main dir and generate the according `Makefile` in the subdir `librna`.

librna/ViennaRNA/static/misc/dna_mathews1999.hex
```diff
@@ -52,10 +52,10 @@
   0x2d, 0x32, 0x30, 0x20, 0x20, 0x20, 0x2d, 0x36, 0x30, 0x20, 0x20, 0x2d,
   0x31, 0x33, 0x30, 0x20, 0x20, 0x2d, 0x31, 0x36, 0x30, 0x20, 0x20, 0x20,
   0x20, 0x20, 0x30, 0x0a, 0x20, 0x20, 0x2d, 0x31, 0x39, 0x30, 0x20, 0x20,
-  0x2d, 0x32, 0x32, 0x30, 0x20, 0x20, 0x20, 0x44, 0x45, 0x46, 0x20, 0x20,
+  0x2d, 0x32, 0x32, 0x30, 0x20, 0x20, 0x20, 0x2d, 0x35, 0x30, 0x20, 0x20,
```
kmaibach, line 55: So this here for example is a change the Vienna team made?

Can anyone explain why they did that? I get that the energy models need to be adjusted - but wouldn't you rename the model to something like "mathews1999.2"? 

Did they mention the reason why they changed it?

fymue: Yes, this is a change that the ViennaRNA team made. These `.hex`-files are generated from the `.par`-files in `gapc/librna/paramfiles/`, which were adjusted in the v2.4.18 release. The ViennaRNA commit message for those changes was: "replaced DEF placeholders in parameter files with their actual value -50" (see Changelog). So the changes in the `.hex`-files are probably coming from those new values. What this means, why they changed the DEF placeholders to -50 and why they didn't rename the model I cannot answer.

sjanssen2: for a reason of change see https://github.com/ViennaRNA/ViennaRNA/issues/115

sjanssen2:
There are few tiny diff left in the hex files:
image

librna/ViennaRNA/static/misc/dna_mathews1999.hex
kmaibach: Ah, this makes sense, thank you!

sjanssen2 (approved):

sjanssen2:
I think it is good to go, but we need a last check and approval from @kmaibach

kmaibach (approved):

PR #171, opened 2023-01-18
underflow in `unsigned int` comparison

sjanssen2:
to my surprise the following grammar could not parse the empty word `""`, despite the `state_E_12 = nil(EMPTY)` non-terminal:
```
grammar gra_wrong uses sig_cm(axiom = state_S_0) {
        tabulated { state_E_12 }
  state_S_0 = 
        silent_transition(state_MP_3) |
        silent_transition(state_E_12)
            # h;
  state_MP_3 = pair_transition(CHAR, state_E_12, CHAR)
             # h;
  state_E_12 = nil(EMPTY)
             # h;
}
```
If we force gapc **not** to tabulate the non-terminal `state_E_12` it works. I first thought, the issue is with the additional guards in the generated table functions in `out.hh`, e.g. 
```
BigInt &  get(unsigned int t_0_i, unsigned int t_0_j) {
   ...
   if ((((t_0_i > 1) || (t_0_j < (t_0_n - 1))) || (t_0_j > (t_0_n - 1)))) {
      return zero;
   }
   ...
}
```
Looks like the `-1` violates the deduced yield size of the non-terminal, which is introduced during table initialization. But this was a wrong assumption.
In fact, the `t_0_j < (t_0_n - 1)` comparison fails if `t_0_n` is 0 like for the empty word. Then, one would have to compare `0 < -1` which does not fit the unsigned int data type and the comparison erroneously was true.
I am changing that here, by adding `+1` to both sides of the operator to avoid negative values.


sjanssen2:
up to here, compile code has not changed, I just wanted to show that the test in fact failes

sjanssen2:
it's not the number `-1` but an underflow with the comparison `t_0_j < t_0_n - 1` for empty inputs, as `t_0_n` is `0` and `0 - 1` is not defined for unsigned int. Here, the comparison results in true which is incorrect.

fymue (commented):
Looks good to me. I'm sure you've considered this, but Just 1 question: 
Since you modified the `Tablegen::offset_const` function to fix this bug, this increment by 1 on both sides of the comparison will now be done for every comparison that gets generated in this function (regardless of which particular `GAPL` program is parsed), right?  If so, could this maybe causes problems in the generated code of other `GAPL` programs/grammars?

fymue:
Yes, mathematically it should be the same thing. And you're right, if this does cause problems elsewhere, the test suite would've probably caught it.

fymue (approved):

kmaibach (approved):

PR #174, opened 2023-01-20
Fix constfloat indices

sjanssen2:


sjanssen2:
closed in favor of #181

PR #176, opened 2023-02-08
"complete_track" ->"minsize(0)" / "maxsize(0)

sjanssen2:
in preparation of yield size computation, I switch from agnostic `complete_track` filter to yield size observing built in `min_size(0) && max_size(0)` for the additional alternative from an outside NT to the inside axiom


sjanssen2:
closed in favor of #181

PR #183, opened 2023-03-01
Gquad

sjanssen2:
This PR shall pave to road for G-Quadruplex integration into RNA folding grammar. We need to
1) expose two new energy functions
2) extend the shape alphabet such that we can represent a quadruplex via `G`


rtlib/shape_alph.hh
```diff
@@ -38,28 +47,34 @@ struct ShapeAlph {
...
+        // set_one(t, l);
+        // set_one(t, l-1);
+        t |= T(4) << (l-(char_width-1));
         break;
       default: assert(false);
     }
   }
   char to_char(T &t, Size i) const {
-    switch (t >> i & T(3)) {
+    switch (t >> i & T(pow(2, static_cast<int>(this->char_width))-1)) {
```
fymue, line 73: Wouldn't it make sense to store the result of the `pow` call in a member variable since its basically a constant? This way the value gets recalculated every time this method is called, which probably won't affect execution time much but is kind of unnecessary nonetheless.

sjanssen2: true. I hope that `gcc`'s optimization routines would apply here, but I am not sure!

fymue: I'm sure `gcc` can optimize something like this, but I have no idea if that's going to happen here. It should be able to detect this though.

sjanssen2: done. Better save than sorry ;-)

fymue (approved):

sjanssen2:
thanks  @fymue !

PR #185, opened 2023-03-03
extend checkpointing mechanism to be `cyk`-compatible

fymue:
By default, `GAPC` uses the top-down Unger-style code generation mechanism to generate the DP algorithms.
In this style, the answer is computed in one giant recursive call. Also, for every non-terminal table an additional `tabulated` vector is generated, which is used to keep track of the table cells that have already been calculated/tabulated once. This information is then read at the beginning of every non-terminal function in an `is_tabulated` check to determine whether or not the current to-be-calculated value has already been tabulated. If that's the case, the value is simply returned immediately. If that is not the case, the value will be calculated and finally added to the table, after which it can be immediately returned whenever it is requested again.

This mechanism is leveraged by the checkpointing mechanism by not only archiving the non-terminal tables but also the `tabulated` vector. If a checkpoint is loaded at the beginning of a program, the information which table values have already been tabulated is also loaded, so all already tabulated values can immediately be returned whenever the DP algorithm tries to calculate these values, which allows the DP algorithm to advance significantly faster and essentially simulates the continuation from the loaded checkpoint.

In the DP algorithms generated in `cyk`-style however, these `is_tabulated` checks are not performed at the beginning of the non-terminal functions. Instead, all table values are calculated in a (nested) loop, which is constructed in a way that ensures that all table values for indices smaller than the current loop index have already been calculated and added to the tables.

This is key, because this essentially makes the current loop indices "progress markers" for the tabulation status of the tables. So by adjusting the checkpointing mechanism to archive the loop indices instead of the (here non-existent) tabulated vectors in `--cyk` mode, we can load the table archives as well as these indices at program startup and simply continue the `cyk` loop from the loaded index values, which again simulates the continuation of the program from the loaded checkpoint.


rtlib/fair_mutex.hh
```diff
@@ -0,0 +1,75 @@
...
+ * this isn't guranteed by std::mutex and could cause a thread to repeatedly
+ * get to reaquire the mutex if there isn't enough time for the OS/thread
+ * library to detect that another thread was also waiting to aquire the mutex
+ */
+
+#ifndef RTLIB_FAIR_MUTEX_HH_
+#define RTLIB_FAIR_MUTEX_HH_
+
+#include <cassert>
+#include <chrono>
```
sjanssen2, line 33: I suggest to append lines with `  // NOLINT [build/c++11]` to suppress this linter warning

src/checkpoint.hh
```diff
@@ -1282,9 +1521,10 @@ SUPPORTED_EXTERNAL_TYPES = {"Rope", "answer_pknot_mfe", "pktype",
   }
 
   void parse_checkpoint_log(Printer::Base &stream) {
-     inc_indent(); inc_indent();
+     inc_indent();
+     if (!cyk) inc_indent();
      stream << indent() << "void parse_checkpoint_log("
```
sjanssen2, line 1526: identation in generated code seems to be off by one

src/cpp.cc
```diff
@@ -2122,16 +2167,49 @@ void Printer::Cpp::multi_print_cyk(
...
+    if (checkpoint) {
+      /*
+         in checkpointing mode, the loop indices (e.g. t_0_i and t_0_j
+         in single-track mode; names can differ in multi-track mode)
+         are members of the out class instead of local loop variables
+         so they can be archived/loaded;
+         if they are loaded, the outer loop variable (e.g. t_0_j) starts
+         from whatever value was loaded from the checkpoint
+      */
+      stream << indent() << "for (; " << js << " < "
```
sjanssen2, line 2184: when running through by debugger it looks like the loop variables are NOT properly initialized?!

`testdata/grammar/adpf.gap -p mfe --cyk --checkpoint`

src/cpp.cc
```diff
@@ -2122,16 +2167,49 @@ void Printer::Cpp::multi_print_cyk(
...
+         in checkpointing mode, the inner loop variable (e.g. t_0_i)
+         is either regularly set to e.g. t_0_j + 1 or to whatever value
+         was loaded from the checkpoint archive;
+         however, this loaded checkpoint value will only be the start value
+         for t_0_i in the first inner loop pass;
+         afterwards it will regularly be set to t_0_j + 1;
+      */
+      stream << indent() << "for (" << is << " = ("
+             << is << " == 1 ? " << js << " + 1 : " << is << "); "
+             << is << " > 1; " << is << "--) {" << endl;
```
sjanssen2, line 2205: will this carry over to multitrack grammars, e.g. https://github.com/jlab/ADP_collection/blob/master/alignments.gap? There you can have #tracks * 2 nested loops

testdata/regresstest/config
```diff
@@ -452,6 +452,13 @@ check_checkpoint_eq adpf.gap unused shape5pfx "CGAGCGUAGUCGUAGCguCGUAGCGUAGUCGUA
...
+# std::pair<int, Rope> + cyk
+check_checkpoint_eq adpf.gap unused mfeen "CGAGCGGUACUGAGUCGUAGUGUAGUGCUCGUAGUCGUAGUCGUAGUCGUAGCUGAUCGUAGUAGUCGUCGUAGUCGUAGUCGUCGUAGCGUAGUCGUAGUCGUAGUCAGUCGUGUCGUGUACGUAGUCGUCAGUGCGAUCGUGUCGUAGUCGUCGUAGUCGUAGUCGAGUCAGUCGUAGUAGUGCUGAUCGUgauCGUAGCGUA" checkpoint9 1
+# std::pair<int, String> + cyk
+check_checkpoint_eq adpf.gap unused mfepp "CGAGCGGUACUGAGUCGUAGUGUAGUGCUCGUAGUCGUAGUCGUAGUCGUAGCUGAUCGUAGUAGUCGUCGUAGUCGUAGUCGUCGUAGCGUAGUCGUAGUCGUAGUCAGUCGUGUCGUGUACGUAGUCGUCAGUGCGAUCGUGUCGUAGUCGUCGUAGUCGUAGUCGAGUCAGUCGUAGUAGUGCUGAUCGUgauCGUAGCGUAGUCGAUCGUGAUGUCGUAGUCGUGUAGCUGAUGCUGUAGUCGUAUGCGUAGUCGUGUACGUGUAGCU" checkpoint10 1
 GAPC="../../../gapc"
```
sjanssen2, line 462: can you please also test multi track grammars like `grammar/ali2.gap`?

src/checkpoint.hh
```diff
@@ -703,6 +730,55 @@ SUPPORTED_EXTERNAL_TYPES = {"Rope", "answer_pknot_mfe", "pktype",
...
+               << "map_1d_to_2d(vec_i);" << endl;
+       stream << indent() << "i = indices.first; j = indices.second;" << endl;
+        stream << indent() << "switch (table_i) {" << endl;
+        inc_indent();
+        int c = 0;
+        for (auto i = tables.begin(); i != tables.end(); ++i) {
+          const std::string &table_name = i->first;
+          stream << indent() << "case " << c << " :" << endl;
+          inc_indent();
+          // get table configuration (quadratic, linear etc.)
```
sjanssen2, line 757: attention: CONST tables can have running indices. This is why Gregor checks for indices being deleted like here https://github.com/jlab/gapc/blob/4bf7b5061136becd2bff70b58c9f5fbc95867ad8/src/cpp.cc#L2041-L2046 I struggled with this for a while for outside computation and finally also added https://github.com/jlab/gapc/blob/9337c750cc42be4c2fddddf90c31891088f4e3c8/src/table.cc#L73-L75

Not sure if this applies here, just wanted to make sure you are aware of these edge cases

src/checkpoint.hh
```diff
@@ -934,6 +1019,124 @@ SUPPORTED_EXTERNAL_TYPES = {"Rope", "answer_pknot_mfe", "pktype",
...
+            << "tmp_out_cyk_path.c_str()"
+            << ", std::ios::binary);" << endl;
+     stream << indent() << "if (!(array_fout.good())) {" << endl;
+     stream << indent() << "  throw std::ofstream::failure(\"\");" << endl;
+     stream << indent() << "}" << endl;
+     stream << indent() << "boost::archive::binary_oarchive "
+                            "array_out(array_fout);" << endl;
+     for (size_t i = 0; i < n_tracks; i++) {
+       stream << indent() << "array_out << t_" << i << "_i + 1;" << endl;
+       stream << indent() << "array_out << t_" << i << "_j - 1;" << endl;
```
sjanssen2, line 1039: why the +/- 1 offset? I figure to ensure that a loop iteration is not half computed. I am thinking about my outside version where indices run into the opposite direction. Can you leave a note here such that I will be reminded when merging outside?

src/checkpoint.hh
```diff
@@ -934,6 +1019,124 @@ SUPPORTED_EXTERNAL_TYPES = {"Rope", "answer_pknot_mfe", "pktype",
...
+                            "trying to read cyk loop indices!\\n\";" << endl;
+     stream << indent() << "std::exit(1);" << endl;
+     dec_indent();
+     stream << indent() << "}" << endl;
+     dec_indent();
+     stream << indent() << "}" << endl << endl;
+     dec_indent();
+  }
+
+  void map_1d_to_2d(Printer::Base &stream) {
```
sjanssen2, line 1120: when is this fct called?

src/cpp.cc
```diff
@@ -1537,6 +1537,18 @@ void Printer::Cpp::print_seq_init(const AST &ast) {
...
     stream << indent() << "logfile_path = opts.checkpoint_out_path / "
            << "logfile_name;" << endl << endl;
+    if (ast.checkpoint->cyk) {
+      stream << indent() << "std::string cyk_archive = "
+             << "file_prefix + \"_cyk_indices\";" << endl;
+      stream << indent() << "out_cyk_path = opts.checkpoint_out_path / "
+             << "cyk_archive;" << endl;
+      stream << indent() << "tmp_out_cyk_path = opts.checkpoint_out_path / "
+             << "(cyk_archive + \"_new\");" << endl << endl;
+      for (size_t i = 0; i < ast.grammar()->axiom->tracks(); i++) {
```
sjanssen2, line 1547: it is not guaranteed that any NT has the same number of tracks as the axiom!

src/cpp.cc
```diff
@@ -1875,6 +1909,17 @@ void Printer::Cpp::header(const AST &ast) {
 
   print_most_decl(*ast.grammar()->axiom);
 
+  if (ast.checkpoint && ast.checkpoint->cyk) {
+    ::Type::Base *type = new Type::Size();
+    // store indices for cyk loops here so they can be archived/loaded
+    stream << indent() << "// indices for cyk loops" << endl;
+    for (size_t t = 0; t < ast.grammar()->axiom->tracks(); t++) {
```
sjanssen2, line 1916: see above, regarding # tracks

sjanssen2 (changes requested):
first round of review - I have to admit that I do not always fully understand the code change consequences, therefore you might find some comment a bit stupid.
General question: will this technique carry over when compile flag `_OPENMP` is set?

rtlib/fair_mutex.hh
fymue: That would work too.

src/cpp.cc
fymue: Thats weird. I'm initializing the loop variables in the `out::init` function, so they should be initialized before the `cyk` function is called.

src/checkpoint.hh
fymue: I added the offset, because you mentioned it might be safer if we don't continue from the exact loop index at the point of the dump but from 1 iteration prior. So I chose to go -1 for the outer loop because the index is incremented and +1 for the inner loop index because it is decremented.

src/checkpoint.hh
fymue: This function will be called in the `restore_string_links` function if the table contains `String` types and we checkpoint in `cyk` mode. Since there is no `tabulated` vector in `cyk` mode, I can't just mark the broken Lists/Strings as not tabulated to force the DP algorithm to recalculate them. Instead I have to manually call the appropriate non-terminal functions to recalculate the broken Lists/Strings. These non-terminal functions take two indices as input though while I only have access to the index of the broken List in its `std::vector`. So I need to map the 1D vector index to a `row` and `column` index as if the 1D vector were a 2D (upper triangular) matrix, which this function does.  

A little complicated, I know.

src/cpp.cc
fymue: Ah, good to know. It's kind of hard to figure all the differences between the different tables just from reading the source code, especially since you've also mentioned that the `CONST` table can also have running indices. I will have another look at the whole mechanism and test some more (multi track) programs to hopefully be able to spot all the edge cases.

testdata/regresstest/config
fymue: Will do.

src/cpp.cc
fymue: It should, but I'm not sure. I will test that.

src/checkpoint.hh
fymue: Fixed.

fymue:
The last two commits should handle multi-track cyk functions properly now.

I also realized that I won't need to delete/recalculate any broken `String` objects in `cyk` mode, because we can only dump if a whole loop iteration has been completed, which ensures that all links can be resolved after deserialization. This means that I was able to delete the `map_1d_to_2d` function and to remove the final part of the `restore_string_links` function.

I also extended the checkpointing tests to be able to test two-track (cyk) programs. I will push the new tests after [this PR](https://github.com/jlab/gapc-test-suite/pull/15) has been merged.

The final feature that's still missing would be the `OPENMP` `cyk` checkpointing implementation, so I will work on that next.

src/checkpoint.hh
fymue: This function is now no longer needed (see last comment).

fymue:
The checkpointing mechanism should now also be able to handle programs compiled with `cyk` + `OPENMP`.

It was a little bit tricky to be able to start the loops from the loaded indices since it wasn't immediately clear to me how the variables behave in mutlithreading /  `OPENMP` since multiple threads are involved, but I managed to figure it out.

I also needed to add a custom `shared_mutex` class, since I needed a mutex that can be (un)locked by multiple threads and that ensures that only **one** thread can own/lock the mutex if needed (i.e. the archiving thread).

Furthermore, I needed to achieve synchronization between all threads at certain points in the loops to ensure e.g. that no "half" iterations are checkpointed, which is why you might see `#pragma omp ordered` in certain sections of the code. 

@sjanssen2 
I guess this is intended, but in order to trigger the `OPENMP` loops I needed to compile the `*.cc`-files with the `-fopenmp` flag. By default, the generated `Makefile` only adds this flag in the `out` compilation command, which doesn't activate `OPENMP`.  
During testing I also noticed that certain products cause runtime errors when you (manually) compile with `OPENMP`. For example, binaries compiled from `gapc -p alg_mfe*alg_dotBracket --cyk nodangle.gap`  or  `gapc -p alg_mfe*alg_enum --cyk nodangle.gap`(with `-fopenmp` added for all `.cc`  files) crash. Maybe that's why `OPENMP` isn't activated by default?

sjanssen2:
@fymue you are right regarding openMP. It was only quick and dirty developed as a proof of principle. It turned out, that the runtime savings did not pay off for the short typical input sequences of that time. Thus, openMP is not trivially activated for end user. I'd leave as is for now

sjanssen2:
I just ask myself if it would make sense to store the CYK indices in plain text, such that the user can a) inspect those with the risk that he/she b) might tinker with them?!

src/checkpoint.hh
```diff
@@ -934,6 +978,117 @@ SUPPORTED_EXTERNAL_TYPES = {"Rope", "answer_pknot_mfe", "pktype",
...
+       // archive the indices of the prior iteration to ensure
+       // that the whole iteration has been completed
+       stream << indent() << "array_out << t_" << i << "_i + 1;" << endl;
+       stream << indent() << "array_out << "
+              << "(t_" << i << "_j == 0 ? 0 : t_" << i << "_j - 1);" << endl;
+     }
+     stream << indent() << "#ifdef _OPENMP" << endl;
+     stream << indent() << "  array_out << z1_;" << endl;
+     stream << indent() << "  array_out << z2_;" << endl;
+     stream << indent() << "  array_out << y_;" << endl;
```
sjanssen2, line 1006: what are those z1_, z2_ and y variables? Did you introduce them or were they already present. Could we rename them into something more speaking?

src/checkpoint.hh
```diff
@@ -994,7 +1149,7 @@ SUPPORTED_EXTERNAL_TYPES = {"Rope", "answer_pknot_mfe", "pktype",
      stream << "double get_tabulated_vals_percentage() const {" << endl;
      inc_indent();
      stream << indent() << "return static_cast<double>(tabulated_vals_counter) "
-            << "/ tabulated.size() * 100.0;" << endl;
```
sjanssen2, line 997: can we format the percentage values to "%.2f", I feel `69.4831%` is too precise, better make it `69.48%` and then more it thus consistent: `Table is 0% filled.`--> `Table is 0.00% filled.`
Btw: I really love this feature :-)

sjanssen2 (approved):
great work: I just ask for one stylistic change (format % table filled) and one semantic question regarding `z_` variables.

fymue:
> I just ask myself if it would make sense to store the CYK indices in plain text, such that the user can a) inspect those with the risk that he/she b) might tinker with them?!

I thought about this as well, but I figured it would be more safe the create a dedicated archive for these indices. I could additionally write them Logfile though so the user can at least inspect them. Should I do that?

src/checkpoint.hh
fymue: Sure.

 I think it's a cool feature as well. Had fun implementing this and learnt a lot!

fymue:
> wouldn't it do to switch from a `binary_oarchive` to a plain text archive stream? If this is simple, I tend to ask you to do so. But storing indices in the archive plus reporting them in the log sounds dangerous, as they might be manipulated asynchronously by users

That would also work. I will do that then.

PR #187, opened 2023-03-07
Outside

sjanssen2:
This PR is huge. Sorry. It adds the ability of automatic outside grammar generation to gapc, via the parameter `--outside_grammar ALL`.

The current implementation can "consume" a subword i,j from the user provided input sequence(s) until every symbol between i and j are parts of candidates (the inside direction). With this PR, we revert the direction: given i,j we "consume" characters 0...i and j...n until we reach 0 and n (the outside direction). In practice, this allows to conceptually split all candidates at a give i,j and compute inside (was already possible) and outside (new) parts which combined will produce complete candidates. With this, we can easily compute e.g. posteriors or other useful information of the candidate space.

I've tried to capsule the code into three phases and create offload most of the new functions into `src/outside`, but of course changes also touch existing inside parts and thus need changes of existing code - which I tried to keep minimal. The phases are
1. `grammar_transformation`: convert the user defined inside grammar into a grammar that additionally contains outside rules which reflect the structure of the inside grammar, but operate from inside to outside
2. `middle_end`: the new direction requires running indices from i,j towards 0 and n. Thus, moving boundaries and loops require reverse order. Code in middle_end produces these different non-terminal functions.
3. `codegen`: result of the algorithm is no longer the single DP cell (0,n) for the axiom, but values of every (tabulated) NT at every position i,j. Thus, codegen produces a function `print_insideoutside_report_fn` to report these many values (also consider multi track grammar with more than two dimensions). To limit the output, a use can define which NTs shall be reported via the gapc parameter `--outside_grammar X` where X is one non-terminal. Repeatedly use of `--outside_grammar` with different X will lead to multiple NTs being reported. If the user provides `ALL` as non-terminal, all NTs will be reported.

We ran multiple semantic checks to warn users if outside grammar generation is not meaningful (e.g. empty word cannot be parsed) or algebra functions would use mixed data types

The new function shall work with CYK, CYK+openMP (only single track) and Unger code generation, checkpointing, multitrack.

- semantic checks:
  - [x] grammar cannot parse the empty word
  - [x] are all requested NTs in the grammar, for `--outside_grammar`?
  - [x] does the algebra use mixed types, i.e. `answer foo(answer_bar)`, if so outside cannot be generated as we will lack `answer_bar foo(answer)` for outside parts
     - [x] complete definition of `is_terminal_type` for ALL types!
- resolve_blocks:
   - [x] check if `multi_filter` really needs to become `public`
- other ToDos
  - [x] remove `#ifdef LOOPDEBUG` parts
  - [x] add documentation for middle_end and codegen
  - [x] don't forget to revert actions back to master branch of gapc-test-suite!
  - [x] cyk
  - [x] cyk + checkpointing
  - [x] cyk + openMP
  - [x] cyk + openMP + checkpointing
  - [x] how to deal with non-terminals that have no choice function applied?

P.S. this PR shall replace #122


fymue:
Wow, this is indeed a huge PR and a lot of changes to go through. I will try to work through them ASAP and will ask questions if any come up. 

One immediate question though: You mentioned that this PR will replace #122 (`outside_looporder`), which is the branch/PR that the derivative PRs #151 and #160 are based on. I depend on those PRs for my work on the PyTorch interface for derivative code (#166). Can we somehow merge the 1st/2nd derivative changes/branches with this `outside` branch instead of the `outside_looporder` branch? I don't know how/if that would even be possible since those PRs are directed at `outside_looporder` and thus depend on the changes on that branch, and we probably can't change the target branch of those PRs without breaking those changes. At worst, we might need to reimplement the `derivative` implementation for the `outside` implementation of this branch, which would obviously be rather annoying.

sjanssen2:
> One immediate question though: You mentioned that this PR will replace #122 (`outside_looporder`), which is the branch/PR that the derivative PRs #151 and #160 are based on. I depend on those PRs for my work on the PyTorch interface for derivative code (#166). Can we somehow merge the 1st/2nd derivative changes/branches with this `outside` branch instead of the `outside_looporder` branch? I don't know how/if that would even be possible since those PRs are directed at `outside_looporder` and thus depend on the changes on that branch, and we probably can't change the target branch of those PRs without breaking those changes. At worst, we might need to reimplement the `derivative` implementation for the `outside` implementation of this branch, which would obviously be rather annoying.

I will take care of this :-) #122 is a proof of principle and rather ugly. I will carry over changes for derivatives into a new PR based on outside and continue from there towards your pytorch branch ... a lot of work, but worth it!

fymue:
> I will take care of this :-) #122 is a proof of principle and rather ugly. I will carry over changes for derivatives into a new PR based on outside and continue from there towards your pytorch branch ... a lot of work, but worth it!

Alright, that'll work. This indeed does sounds like a lot of work though. I hope carrying over those changes won't be too difficult.

src/fn_arg.hh
```diff
@@ -144,6 +149,21 @@ class Base {
  public:
     virtual void init_multi_ys() = 0;
     virtual const Yield::Multi &multi_ys() const { return m_ys; }
+
+    const bool is_partof_outside() {
```
fymue, line 153: Shouldn't this be `bool is_partof_outside() const {`? 

Returning `const bool` would indicate that the boolean returned from this function is constant, i.e. doesn't change. Since there is a `set_partof_outside` method though, I would assume that this boolean can be changed at runtime, which I think can be confusing if there is a `const bool` method for the same boolean.  `bool is_partof_outside() const`, however, just means that this method doesn't change the internal state of the object is belongs to, which applies here.

src/grammar.hh
```diff
@@ -213,6 +216,35 @@ class Grammar {
...
+   * Throws an error if NTs are not present.
+   */
+  void check_outside_requested_nonexisting_nts();
+
+  /* not yet complete: will inject additional production rules such that
+   * the given (inside-) grammar is turned into an outside grammar.
+   */
+  void convert_to_outside();
+
+  const bool is_partof_outside() {
```
fymue, line 235: Same thing as in previous comment.

src/fn_arg.hh
fymue: Yes, that's what I thought. It's quite easy to accidentally put the `const` in the wrong place. I found the same issue in `grammar.hh` (see below), so maybe this will also appear in some other files, in which case I won't point it out again.

src/outside/grammar_transformation.cc
```diff
@@ -0,0 +1,967 @@
...
+   */
+
+  if (!this->ast.get_outside_nt_list()) {
+    // the user did not request any outside NT to be reported
+    return;
+  }
+  assert(this->ast.get_outside_nt_list());
+
+  // a list that collects the names of non existent NTs
+  std::list<std::string> *warn_missing_nts = new std::list<std::string>();
```
fymue, line 63: The `warn_missing_nts` vector is not properly deleted at the end the function, creating a small memory leak. You could either just stack-allocate it (I don't see a reason why this needs to be heap-allocated with `new` here) or add `delete warn_missing_nts` at the end of the function to avoid this. 

Or we let the OS clean up the memory after GAP-C exits, which we rely on anyways in other areas of the compiler, but obviously isn't really the proper way to clean up memory.

src/outside/grammar_transformation.cc
```diff
@@ -0,0 +1,967 @@
...
+      t->is(Type::TABLE) ||  // a DP matrix to store non-terminal results
+      t->is(Type::LIST) ||  // the answer list of an algebra function
+      t->is(Type::TUPLE) ||  // a pair of results
+      t->is(Type::CHOICE) ||  // flag an algebra function as a choice function
+      t->is(Type::RANGE) ||  // codegen, set of begin/end iterators
+      t->is(Type::SIZE)  // internal type for code generation = unsigned int
+      ) {
+    return false;
+  }
+  if ((dynamic_cast<Type::Name*>(t)) ||  // the name of a new user defined tuple
```
fymue, line 164: I don't really understand the purpose of this `if` statement. Dynamic-casting the `Type::Base* t` to pointers to derived classes of this base type and checking them like this in an `if` statement will always evaluate to `true` if `t != nullptr`, regardless of the type `t` is casted into. If  `t == nullptr`, all casts inside the `if` statement would evaluate to `false`. If a certain cast wouldn't be allowed/possible, `dynamic_cast` would cause a compiler error. So essentially this `if` statement is just checking if  `t != nullptr` (or just `if (t)` simplified) , which I think should either be the first check inside `is_terminal_type` so we don't try to dereference a null pointer here, or just left out completely if we can be sure that `t` is never going to be a null pointer.

src/outside/grammar_transformation.cc
sjanssen2: I'd prefer to delete at the end ... because debugging via Eclipse / gdb often stalls with stack-allocations (not sure why)

src/grammar.hh
sjanssen2: I went through the code changes and hopefully resolved all these situations.

src/outside/grammar_transformation.cc
fymue: My mistake. `dynamic_cast` actually does return `nullptr` if a cast from a Base class to derived class fails due to the derived class being the wrong type. I wrongly assumed that this would also cause an error. So this `if` statement now does make sense and can be used to figure out if `t` is of a specific derived class.

However, you're also rightfully pointed out that `Type::Name` isn't a derived type from `Type::Base`. So for that type this cast probably doesn't work as intended as causes undefined behaviour according to [cppreference](https://en.cppreference.com/w/cpp/language/dynamic_cast). So here we would either need to make `Type::Name` a derived type of `Type::Base` or remove the check for `Type::Name` in the `if` statement.

src/outside/grammar_transformation.cc
fymue: That would be perfectly fine as well.

fymue:
@sjanssen2 General question: Are you generally OK with (small) memory leaks in the compiler? I'm asking because I found quite a few memory leaks in `grammar_transformation.cc`, which I can't all point out here without spamming the comment section, but also in other parts of the compiler.

This is a pretty subjective question, but in general we obviously want to avoid memory leaks in software to a) not use up unnecessary memory and b) to avoid security issues. Both of those things don't really matter in GAP-C, because it doesn't need much memory to transpile GAP-L code, and also isn't software that is subject to attacks, so security is secondary. What's your opinion on this? If you are OK with minor memory leaks, I won't point them out again.

sjanssen2:
I totally agree with you, we should avoid them. This is just due to my poor C++ programming skills. Please spam as much as necessary :-)

src/outside/codegen.cc
```diff
@@ -0,0 +1,229 @@
...
+
+  return report_nts;
+}
+
+
+void print_insideoutside_report_fn(Printer::Cpp &stream, const AST &ast) {
+  bool old_in_class = stream.in_class;
+  stream.in_class = true;
+
+  std::list<Symbol::NT*> *report_nts = NTs_to_report(ast);
```
fymue, line 101: `report_nts` should be deleted at the end of the `print_insideoutside_report_fn` function.

fymue: Actually, if I'm not mistaken, the code for the `report_insideoutside` function is directly added to the output file(stream) at the end of `print_insideoutside_report_fn`, right? In that case, literally every heap-allocated object can be deleted at the end of the function. I definitely can't point all of those out individually. If you want I can try to adjust this function myself to properly delete everything at the end. Shouldn't take very long, its just too much work to point every individual example out.

src/outside/grammar_transformation.cc
```diff
@@ -0,0 +1,969 @@
...
+};
+
+
+/* iterates through the rhs alternatives of an NT and creates a clone of an
+ * alternative where ONE (but not all) rhs NT is swapped with the lhs NT, e.g.
+ *   struct = cadd(dangle, weak) | sadd(BASE, struct) will result in
+ * a) outside_dangle = cadd(outside_struct, weak)
+ * b) outside_weak = cadd(dangle, outside_struct)
+ * c) outside_struct = sadd(BASE, outside_struct) */
+struct Flip_lhs_rhs_nonterminals : public Visitor {
```
fymue, line 530: The objects that get allocated in this class can be deleted when an object of this class gets destroyed:
```
~Flip_lhs_rhs_nonterminals() {
    delete alt_clones;
    delete lhs_nt->name;
    delete lhs_nt;
  }
```

src/outside/codegen.cc
sjanssen2: your assumption is correct, `print_insideoutside_report_fn` produces the code that reports the DP cell values after inside and outside pass. Your contribution is highly appreciated!

src/outside/grammar_transformation.cc
sjanssen2: why do we need to delete `lhs_nt->name` if we delete `lhs_nt` in the line below?

fymue: Because the `NT` object doesn't free the memory of the `name` ptr (or any other ptr) it stores when it gets destructed. So after destruction of the `NT`, only the ptr to `name` will be deleted, but no the actual data `name` pointed to.

fymue: Looks like trying to free the memory in the destructors doesn't work as intended. I scanned the code a bit more and think that the data that the ptrs of this object point to is used after the object has already been destroyed (e.g. `convert_to_outside` in `grammar_transformation.cc`, lines 886ff.), which can obviously lead to problems if we've deleted that data already. I guess we will have to live with not freeing that memory here.

src/grammar.cc
```diff
@@ -655,7 +673,11 @@ void Grammar::put_table_conf(std::ostream &s) {
       hashtable<std::string, Symbol::NT*>::iterator a = tabulated.find(
         *i->second->name);
       if (a == tabulated.end()) {
-        assert(false);
+        // user might not have considered outside NTs to be tabulated in
```
kmaibach, line 676: does the automated table design consider the NTs of the outside grammar as well when calculating a good design?

src/outside/codegen.cc
```diff
@@ -0,0 +1,231 @@
...
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+}}} */
+
+#include "codegen.hh"
+
+
+std::list<Symbol::NT*> *NTs_to_report(const AST &ast) {
+  /* define which non-terminals shell be reported to the user
```
kmaibach, line 28: 'shall' not 'shell'

kmaibach (commented):
This is indeed a very large PR and I fear that I will not understand everything that's going on here. But I try to work through it and hope it's fine that I'll ask some questions while doing so.

1) One general question regarding grammar transformation: 
Can the grammar be saved to a file or is it only reported to stdout?
Would it be useful to include that so the user can have a look at it later?

2) Just for my understanding: Why would I want to exclude NTs from outside transformation? Are there cases where some NTs don't influence the probabilities of the rest of a grammar? Or are those only NTs of the form X -> a ?

3) Not necessary for this PR but out of curiosity: Did you check the runtime of the code generation regarding grammar size (number of NTs)? Would be interestring how it changes with larger grammars.

src/grammar.cc
sjanssen2: yes it does :-) Watch out for `outside_multiloop` in the following example. A heuristic table design `-t` would not tabulate this NT, while forcing to tabulate all NTs will of course also create a table for `outside_multiloop`:
```
(cpplint) t490s x86_64 /Daten/Git/jlab/gapc_backward>./gapc testdata/grammar_outside/nodangle.gap --outside_grammar ALL -l 0 -t
Info: 
Grammar has been modified into an outside version.
The following non-terminals will be tabulated (%name indicates linear table, %name% indicates constant tables):
  {dangle, iloop, ml_comps, ml_comps1, outside_dangle, outside_iloop, outside_ml_comps, outside_ml_comps1, outside_strong, outside_struct, outside_weak, strong, struct, weak, } #14 of 25 

resulting in an estimated runtime of:
  22n^3  + 8956n^2  + 2
(cpplint) t490s x86_64 /Daten/Git/jlab/gapc_backward>./gapc testdata/grammar_outside/nodangle.gap --outside_grammar ALL -l 0 --tab-all
Info: 
Grammar has been modified into an outside version.
The following non-terminals will be tabulated (%name indicates linear table, %name% indicates constant tables):
  {dangle, hairpin, iloop, leftB, ml_comps, ml_comps1, multiloop, %outside_axioms%, outside_dangle, outside_hairpin, outside_iloop, outside_leftB, outside_ml_comps, outside_ml_comps1, outside_multiloop, outside_rightB, outside_stack, outside_strong, outside_struct, outside_weak, rightB, stack, strong, struct, weak, } #25 of 25 

resulting in an estimated runtime of:
  22n^3  + 8966n^2  + 4
(cpplint) t490s x86_64 /Daten/Git/jlab/gapc_backward>
```

sjanssen2:
> 1. One general question regarding grammar transformation:
>        Can the grammar be saved to a file or is it only reported to stdout?
>        Would it be useful to include that so the user can have a look at it later?

The grammar is NOT printed to stdout, but you can "plot" it via `--plot-grammar 1`. This will not report `gap-L` code but a visual representation. And yes, I think it is useful to inspect this automatic generation. I've added this hint to the help message of gapc.

sjanssen2:
> 2\. Just for my understanding: Why would I want to exclude NTs from outside transformation? Are there cases where some NTs don't influence the probabilities of the rest of a grammar? Or are those only NTs of the form X -> a ?

You would not manually consider excluding NTs, but yes, there are "inside" productions that lack outside analogues because they have no r.h.s. non-terminals as in your example! (I guess `a` is a terminal)

sjanssen2:
> 3\. Not necessary for this PR but out of curiosity: Did you check the runtime of the code generation regarding grammar size (number of NTs)? Would be interestring how it changes with larger grammars.

Theory says that every CFG can be transformed into Chomsky Normal Form (CNF), i.e. max. width is 1 and would have at most 2 outside productions. Thus, the asymptotic run time will not change; only a constant factor is added. For CNF, this yields a factor of ~3. For ADP CFGs with width > 1, this factor can be higher (depends on the number of r.h.s. NTs) but the asymptotic remains the same.

sjanssen2:
> hope it's fine that I'll ask some questions

these are great questions! Please don't hesitate to ask more of those!

sjanssen2:
> > > 2. Just for my understanding: Why would I want to exclude NTs from outside transformation? Are there cases where some NTs don't influence the probabilities of the rest of a grammar? Or are those only NTs of the form X -> a ?
> > 
> > 
> > You would not manually consider excluding NTs, but yes, there are "inside" productions that lack outside analogues because they have no r.h.s. non-terminals as in your example! (I guess `a` is a terminal)
> 
> Yes, a is a terminal.
> 
> I think I have seen it somewhere but are there checks to see if an inclusion or exclusion of certain NTs is problematic? Does the user get a warning if they don't include NTs?

outside grammar generation is fully automatic, i.e. the user has no saying in which NTs to process. Therefore, he/she cannot do anything wrong here, except designing an inside grammar that cannot parse the empty word - an according warning will be reported to the user.

kmaibach:
> > > > 2. Just for my understanding: Why would I want to exclude NTs from outside transformation? Are there cases where some NTs don't influence the probabilities of the rest of a grammar? Or are those only NTs of the form X -> a ?
> > > 
> > > 
> > > You would not manually consider excluding NTs, but yes, there are "inside" productions that lack outside analogues because they have no r.h.s. non-terminals as in your example! (I guess `a` is a terminal)
> > 
> > 
> > Yes, a is a terminal.
> > I think I have seen it somewhere but are there checks to see if an inclusion or exclusion of certain NTs is problematic? Does the user get a warning if they don't include NTs?
> 
> outside grammar generation is fully automatic, i.e. the user has no saying in which NTs to process. Therefore, he/she cannot do anything wrong here, except designing an inside grammar that cannot parse the empty word - an according warning will be reported to the user.

I think I completely misunderstood this passage in your explanation:
> To limit the output, a use can define which NTs shall be reported via the gapc parameter --outside_grammar X where X is one non-terminal. Repeatedly use of --outside_grammar with different X will lead to multiple NTs being reported. If the user provides ALL as non-terminal, all NTs will be reported.

I thought that you could exclude the NTs in the code generation NOT (only) in die visual report. But I think the latter is what you really mean, right?

So, my question would be irrelevant then. Sorry for the confusion.

sjanssen2:
> > > > > 2. Just for my understanding: Why would I want to exclude NTs from outside transformation? Are there cases where some NTs don't influence the probabilities of the rest of a grammar? Or are those only NTs of the form X -> a ?
> > > > 
> > > > 
> > > > You would not manually consider excluding NTs, but yes, there are "inside" productions that lack outside analogues because they have no r.h.s. non-terminals as in your example! (I guess `a` is a terminal)
> > > 
> > > 
> > > Yes, a is a terminal.
> > > I think I have seen it somewhere but are there checks to see if an inclusion or exclusion of certain NTs is problematic? Does the user get a warning if they don't include NTs?
> > 
> > 
> > outside grammar generation is fully automatic, i.e. the user has no saying in which NTs to process. Therefore, he/she cannot do anything wrong here, except designing an inside grammar that cannot parse the empty word - an according warning will be reported to the user.
> 
> I think I completely misunderstood this passage in your explanation:
> 
> > To limit the output, a use can define which NTs shall be reported via the gapc parameter --outside_grammar X where X is one non-terminal. Repeatedly use of --outside_grammar with different X will lead to multiple NTs being reported. If the user provides ALL as non-terminal, all NTs will be reported.
> 
> I thought that you could exclude the NTs in the code generation NOT (only) in die visual report. But I think the latter is what you really mean, right?
> 
> So, my question would be irrelevant then. Sorry for the confusion.

ah, that is the misunderstanding. Correct, it only controls which results are reported on stdout. One is typically not interested in all cell values of all NTs, thus one might save printing lines. But in general, we cannot know what the user is interested in.

fymue:
@sjanssen2 Regarding the checkpointing test that keeps failing: I can't really tell what exactly is going wrong by just looking at the log, but I assume that the test input is simply too short for this test if it is executed on multiple threads (with OMP). I ran into some issues before whenever I checkpointed programs that had execution times close to 1s, which is the checkpointing interval used in all tests. We could maybe increase the input length a bit so the test runs a bit longer and see if it still keeps failing.

sjanssen2:
I've increased test input sequences as suggested by @fymue - seems to help :-)
Do you have further issues regarding this PR @fymue @kmaibach ?

fymue:
> I've increased test input sequences as suggested by @fymue - seems to help :-) 

Very good.

>Do you have further issues regarding this PR @fymue @kmaibach ?

Nothing else from my side.

fymue (approved):

testdata/grammar_outside/elmamun.gap
```diff
@@ -0,0 +1,203 @@
...
+    return 0;
+  }
+  choice [int] h([int] candidates) {
+    return list(minimum(candidates));
+  }
+}
+
+grammar gra_elmamun uses sig_elmamun(axiom = formula) {
+  formula = number(INT)
+	  | add(formula, CHAR, formula)
```
kmaibach, line 187: did you forget the '+' here?

testdata/grammar_outside/elmamun.gap
```diff
@@ -0,0 +1,203 @@
...
+	  | mult(formula, CHAR('*'), formula)
+	  | heinz(formula, ROPE("manfred"), formula)
+	  | nil(EMPTY)
```
kmaibach, line 190: you have the algebra function "minus" but did not define it here

testdata/grammar_outside/elmamun.gap
```diff
@@ -0,0 +1,203 @@
...
+  choice [int] h([int] candidates) {
+    return list(minimum(candidates));
+  }
+}
+
+grammar gra_elmamun uses sig_elmamun(axiom = formula) {
+  formula = number(INT)
+	  | add(formula, CHAR, formula)
+	  | mult(formula, CHAR('*'), formula)
+	  | heinz(formula, ROPE("manfred"), formula)
```
kmaibach, line 189: what exactly do you want to test here with this new function? :D

testdata/grammar_outside/elmamun_derivatives.gap
```diff
@@ -0,0 +1,73 @@
...
+    return list(sum(candidates));
+  }
+}
+
+
+grammar gra_elmamun uses sig_elmamun(axiom = formula) {
+  formula = number(INT)
+	  | add(formula, CHAR('+'), formula)
+	  | mult(formula, CHAR('*'), formula)
+	  | nil(EMPTY)
```
kmaibach, line 68: `heinz` and `minus` are missing here

testdata/grammar_outside/mini_complexIL.gap
```diff
@@ -0,0 +1,60 @@
...
+  }
+  int il(Subsequence lb, Subsequence lr, int x, Subsequence rr, Subsequence rb) {
+    return x + il_energy(lr, rr);
+  }
+  int comp(Subsequence a, Subsequence lr,  Subsequence lr2, Subsequence lr3,  int x, Subsequence b, int y) {
+    return x+y;
+  }
+}
+
+//algebra alg_dotBracket implements sig_foldrna(alphabet = char, answer = string) {
```
kmaibach, line 30: multi-line comments would be better

kmaibach (commented):
Alright, so here are my final comments. 
I looked through everything and have some remarks and questions left.

But other than that it should be it.

src/alt.cc
```diff
@@ -1068,25 +1070,58 @@ void Alt::Simple::reset() {
...
+  const Yield::Size &ys, const Yield::Size &lhs, const Yield::Size &rhs,
+  std::list<Statement::For*> &loops,
+  bool for_outside_generation, bool outmost, bool is_left_not_right) {
+  /* only for outside NTs:
+   * If left/right of the rhs outside NT are grammar components with at least
+   * one moving boundary, we need to start leftmost/rightmost with a new
+   * loop variable t_x_k.
+   * Note that empty grammar components might occur on left/rightmost positions
+   * like LOC. Once we encounter these components, we already did increase k
+   * previously. Thus, it is not sufficient to "just" look at lhs, ys and rhs.
```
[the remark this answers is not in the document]
sjanssen2: I inherited these variables names from Georg. They address yield sizes, namely
  - lhs = the combined yield size of parsers left of the current parser
  - rhs = the combined yield size of parsers right of the current parser
  - ys = the yield size (ys) of the current parser
Thus lhs and rhs is kind of correct, but not meant to be within an equation context

src/grammar.hh
```diff
@@ -213,6 +216,35 @@ class Grammar {
...
 
   unsigned int to_dot(unsigned int *nodeID, std::ostream &out, int plot_level);
+
+  /* semantic check: grammar must be able to parse the empty input in
+   * order to provide recursion base for outside candidates */
+  bool check_outside_parse_empty_word();
+
+  /* input check: tests if all NTs requested by the user to be reported
+   * actually do exist in the chosen inside grammar.
+   * Throws an error if NTs are not present.
```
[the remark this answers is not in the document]
sjanssen2: Here, the program will report an "Error" not only a warning and exit. In principle, compilation could continue, but I decided to point the user to the fact that some of his/her intended NTs will never show up in the generated binaries output. Thus, he/she has to intentionally change the compiler call.

testdata/gapc_filter/RFmini_data.hh
```diff
@@ -0,0 +1,83 @@
...
+
+inline int charToIndex(char a) {
+  if ((a == 'A') || (a == 'a')) return 0;
+  if ((a == 'C') || (a == 'c')) return 1;
+  if ((a == 'G') || (a == 'g')) return 2;
+  if ((a == 'U') || (a == 'u')) return 3;
+  return 99;
+}
+inline int charToIndex(char a, char b) {
+  if (((a == 'A') || (a == 'a')) && ((b == 'A') || (b == 'a'))) return 0;
```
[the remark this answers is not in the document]
sjanssen2: this is a very ad hoc GAPC implementation of covariance models. I am sure there is much room for improvement also regarding runtime efficiency. Here, I just want to check correctness for short examples. Should be OK, as this specific file is not part of the compiler - only here for testing.

testdata/gapc_filter/ext_hmm.hh
```diff
@@ -0,0 +1,102 @@
...
+////	  return true; //x.second > 0.3;
+////	  return (x.second > thresh) || (x.second == 1.0);
+////	  //return x.second > 0.5;
+//  }
+//};
+//
+//template <typename SHAPE>
+//inline double getPfuncValue(std::pair<SHAPE, answer_macrostate_pfunc> x) {
+//	return sum_elems(x.second.pf);
+//}
```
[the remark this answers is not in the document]
sjanssen2: this is work in progress and here for testing purpose only. Same as above: not part of the compiler. Read it as a wired user input ;-)
I've removed this part from the file and reformatted the rest for better readability.

testdata/gapc_filter/isntimes.hh
```diff
@@ -0,0 +1,30 @@
...
+  
+  return true;
+}
+
+//a filter to force that a parsed character corresponds to a nucleotide (and not a gap or some structure parts)
+template<typename T>
+inline bool isAnyBase(const Basic_Sequence<> &seq, T i, T j) {
+  if (j<i) return false;
+
+  return ((seq[i] == 'A')	|| (seq[i] == 'C') || (seq[i] == 'G') || (seq[i] == 'U') || (seq[i] == 'R') || (seq[i] == 'Y') || (seq[i] == 'M') || (seq[i] == 'K') || (seq[i] == 'W') || (seq[i] == 'B') || (seq[i] == 'D') || (seq[i] == 'H') || (seq[i] == 'V') || (seq[i] == 'N') || (seq[i] == 'S'));
```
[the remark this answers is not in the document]
sjanssen2: I never call this filter in any test cases, thus I removed the complete function

testdata/grammar_outside/elmamun.gap
sjanssen2: not really. This is a test case and no productive code. Here, I wanted to see why the terminal parser CHAR could be specialized to '*' but ROPE was not able to only parse "manfred", see https://github.com/jlab/gapc/blob/63bd6bf6dc5b220c613e874fd0e9f7ebe3ff3fe9/testdata/regresstest/config#L586-L587

I hope nobody erroneously uses this file for "productive" purposes. Please refer to https://github.com/jlab/ADP_collection/blob/master/elmamun.gap instead!

testdata/grammar_outside/elmamun.gap
sjanssen2: yepp. Legal but unintentionally. But this file was really just used to test `ROPE("manfred")`

testdata/grammar_outside/elmamun.gap
sjanssen2: internals of gapc :-) See https://github.com/jlab/gapc/pull/146/files for further details

testdata/grammar_outside/elmamun_derivatives.gap
sjanssen2: interestingly, the compiler does not complain if an algebra defines the body of an algebra function despite the fact that this algebra function is not declared in the signature! Bug or feature? I am using the behavior to smuggle in a normalization algebra function for computation of derivatives, part of https://github.com/jlab/gapc/pull/151 
Here, I was testing that gapc really does not throw an error about `heinz`.

`minus` is not used in the grammar as this would violate Bellman's Principle in combination with the defined algebra functions. It's basically a left over from copy and pasting the code from the teaching example.

testdata/grammar_outside/mini_complexIL.gap
sjanssen2: or remove this unused algebra definition altogether :-)

sjanssen2:
Hi @kmaibach I hope I have addressed your latest issues appropriately. Can you check again?

src/alt.cc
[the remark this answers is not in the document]
kmaibach: Alright, this makes more sense spelled out, thank you!

kmaibach (approved):

PR #189, opened 2023-03-15
proper align non terminals in grammar plots

sjanssen2:
In current grammar plots `--plot-grammar 1`, the alignment of the non-terminals, which are placed below each other, is not properly left aligned, e.g.
image

This PR uses a different mechanism to inject invisible nodes to force the layout of `dot` to almost perfectly left align all `subgraphs`. We therefore place only one invisible node at the lowest rank per non-terminal, connect this node a) with the lhs non-terminal node and b) with the next lhs non-terminal of the following subgraph. It is important to make the shape rectangular and use proper `ports` for the edge connection, i.e. `sw` to `nw`. We therefore have to return the nodeID of one of the lowest nodes per non-terminal subgraph in addition to the already returned `max_depth`. The result looks visually more pleasing:

image

However, these changes will require updates of all "truth" files in the gapc-test-suite

### DIFF
--- a/src/plot_grammar.cc
+++ b/src/plot_grammar.cc
@@ -38,6 +38,20 @@ static const char *COLOR_NONTERMINAL = "black";
 static const char *COLOR_BLOCK = "gray";
 static const char *COLOR_EVALFCT = "purple";
 
+static const bool HIDEINVISIBLE = true;
+
+std::string make_insivible(bool singlearg) {
+  if (!HIDEINVISIBLE) {
+    return "";
+  } else {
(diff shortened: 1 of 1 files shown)


fymue (approved):
Looks good to me. But I assume you mean that all plot-grammar-related Truth files need to be updated, not literally every Truth file in the gapc-test-suite repo, right? That would be a lot of work..

sjanssen2:
Thanks for your review @fymue I really "just" meant the plot-grammar related truth files. Sorry for the shock ;-)

PR #199, opened 2023-04-21
set yield size of CONST_ terminal parser to (0,0)

sjanssen2:
Production rules in grammars can use two types of special terminal parser. The first is a terminal parser specialized to certain inputs, e.g. `CHAR` can be specialized to just accept a letter `A` like `CHAR('A')`. Yield of this terminal parser is (1, 1).

The second type of special terminal parser is a construct that not really parses parts of the input but is handy to inject parameters to algebra functions, like `foo(CONST_CHAR('x'), ...)`. One might use algebra function `foo` in different rules in the algebra with different constants, another rule could be `foo(CONST_CHAR('y'), ...)`. These terminal parser must have a yield size of (0, 0) as the do not advance in parsing inputs.

Currently, the yield of the `Fn_Arg::Const` was computed incorrectly, i.e. second type did get a non zero yield size. This PR shall fix this issue in preparation of proper summing yield sizes for outside grammar computation.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -31,7 +31,7 @@ jobs:
       run: sudo apt-get install flex bison make libboost-all-dev libgsl-dev python3 python3-pip
 
     - name: Checkout truth
-      run: git clone --branch master https://github.com/jlab/gapc-test-suite.git $GITHUB_WORKSPACE/../gapc-test-suite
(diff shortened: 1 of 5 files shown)


sjanssen2:
@sjanssen2 don't forget to revert truth branch to`master` when merged!

fymue (approved):
Looks good to me.

The Truth branch in the workflow file still has to be reset to `master`, but I see you set yourself a reminder for that already.

PR #204, opened 2023-05-12
Refactor CYK code generation

sjanssen2:
The current code generation for CYK style parsing (in file src/cpp.cc) was directly dumping arbitrary code to the printer stream. The refactoring of this PR lifts code generation into the backend of gapc by using appropriate `Statement` and `Expr`esssion objects. This clears up the recursive schema for multi track single thread code generation and fixes some bugs for checkpoint integration. New functions are collected in the file src/cyk.cc.
I had to add one more Statement class (CustumeCode) to support very specific function calls, e.g. std::mutex locks that would have otherwise caused to much overhead. Use this new Function wisely and avoid wherever possible!

Note to myself: revert truth branch to master, once merged


sjanssen2:
> Is `CustomeCode` something you named intentionally or do you mean "customs" or "custom" (as in `individuell/maßgeschneidert`)?

@kmaibach you got me! This was indeed a typo.

kmaibach:
> > Is `CustomeCode` something you named intentionally or do you mean "customs" or "custom" (as in `individuell/maßgeschneidert`)?
> 
> @kmaibach you got me! This was indeed a typo.

Well, sadly it has to be edited in every file you changed because you were very consistent :D

sjanssen2:
> > > Is `CustomeCode` something you named intentionally or do you mean "customs" or "custom" (as in `individuell/maßgeschneidert`)?
> > 
> > 
> > @kmaibach you got me! This was indeed a typo.
> 
> Well, sadly it has to be edited in every file you changed because you were very consistent : Edit: Ah, I see you've already taken care of that (:

but forgot to rename the variable `CUSTOMECODE` :-/

fymue:
Looks good to me. The comments in `cyk.cc` are a great visualization of how the nested loops actually work, very helpful. 

However, regarding this:

>  The refactoring of this PR lifts code generation into the backend of gapc by using appropriate Statement and Expression objects. This clears up the recursive schema for multi track single thread code generation and fixes some bugs for checkpoint integration.

Did you find actual bugs in the checkpointing mechanism or does this sentence mean that there were some general bugs in the cyk code generation that also affected the checkpointing integration?

sjanssen2:
> Did you find actual bugs in the checkpointing mechanism or does this sentence mean that there were some general bugs in the cyk code generation that also affected the checkpointing integration?

I found some oddities in checkpointing with multi track grammars. There, it looked like the checkpoint dumping worked but loading started from the wrong index. This is most likely due to the unreadable code generation. Let's discuss this offline ... it even gets more complicated if we want to integrate outside grammar parts :-)

fymue:
> I found some oddities in checkpointing with multi track grammars. There, it looked like the checkpoint dumping worked but loading started from the wrong index. This is most likely due to the unreadable code generation. Let's discuss this offline ... it even gets more complicated if we want to integrate outside grammar parts :-)

Alright. I remember trying to properly integrate the checkpointing routine in multitrack cyk programs was quite tedious, so I wouldn't be surprised if I messed something up there.

fymue (approved):

kmaibach (approved):

fymue:
> @fymue the remaining failing test unfortunately also fails on Ubuntu on longer inputs. The issue is with loading a checkpoint in openMP parallel computation. Could you maybe take a look at the generated code example and help me pinpointing the issue?

Sure, I will have a look.

sjanssen2:
using the nomenclature of the documentation of function `cyk_traversal_multithread_parallel`, I suspect that we currently start the loops of the `C` part with the same stored index as the loops in the starting `A` part. Do we have to duplicate indices we store for the checkpoint?

sjanssen2:
Hi @fymue - I am meanwhile under the impression that my previous comment about the wrong start in the `C` part is not correct, as `A` and `B` use the openMP specific cyk indices, while the serial part `C` uses the singlethreaded indices - should be good.

Regarding your question to `max_tiles_n`: I currently don't see a reason why the `tile_size` variable is treated in the special way to be able to provide its value as environment variable. Can't it simply be a cmd arg of the generated program, i.e. make it a member of `opts` through the `generic_opts.hh` mechanism? Once we know this value, `max_tiles_n` should be computed as the division without rest of `t_0_seq.size() \ tile_size`. 

Just to clarify: your current commit does not yet contain this last fix? If so, I wonder why all tests pass - I guess it is due to too short runtimes; since dumping and loading checkpoints for much larger inputs give different values (`2.35419e+307` vs `0` (the empty value)) on my machine.

src/checkpoint.hh
```diff
@@ -1007,8 +1007,22 @@ SUPPORTED_EXTERNAL_TYPES = {"Rope", "answer_pknot_mfe", "pktype",
...
+
+     /*
+      * need to subtract the tile size from inner loop idx, because it is
+      * always incremented by the tile size one too many times in the
+      * parallel OpenMP loop; due to the way the parallelization works,
+      * simply subtracting the tile size once before archiving the index
+      * is the easiest workaround for this
+      */
+     stream << indent() << "#ifdef TILE_SIZE" << endl;
+     stream << indent() << "inner_loop_2_idx -= TILE_SIZE;" << endl;
```
sjanssen2, line 1019: this looks rather ugly. Isn't there a nice way to subtract -1 (times tile_size) in the cyk.cc code?

fymue: There probably is. I will try to clean this up.

fymue:
> Regarding your question to `max_tiles_n`: I currently don't see a reason why the `tile_size` variable is treated in the special way to be able to provide its value as environment variable. Can't it simply be a cmd arg of the generated program, i.e. make it a member of `opts` through the `generic_opts.hh` mechanism? Once we know this value, `max_tiles_n` should be computed as the division without rest of `t_0_seq.size() \ tile_size`.
 
That sounds reasonable. I will implement this and let you take a look at it.

> Just to clarify: your current commit does not yet contain this last fix? If so, I wonder why all tests pass - I guess it is due to too short runtimes; since dumping and loading checkpoints for much larger inputs give different values (`2.35419e+307` vs `0` (the empty value)) on my machine.

I am also a little bit confused why the test passed with just this fix. When it tested it locally on my machine I also needed to adjust the `max_tiles_n` part.

src/checkpoint.hh
fymue: Fixed this with latest commit.

fymue:
I moved the `max_tiles` and `max_tiles_n` computation to the `init` method of the main `out` class. This allows me to initialize the loop indices for the checkpointing mechanism correctly, which should fix the 2nd bug (see earlier comment).

I simoly used the regular print stream to insert the tile computation code into the `init` method for now. With some refactors to `cyk.cc`, we could maybe reuse the `get_tile_computation` function here instead.

fymue:
I refactored the some of the functions in `cyk.cc`  a bit to be able to reuse the `get_tile_computation` function in `cpp.cc` to print the tile computation code in `init`. This means that the tile computation code now no longer needs to be printed in the `cyk` function. Everything else should've remained the same.

Sidenote: I'm pretty sure that the Truth file for the last cyk test that keeps failing is wrong. I recompiled the test program of that test from master and get the answer 3.13527e+163 instead of 2.70468e+162 from the Truth file. We should probably double-check that.

sjanssen2:
@fymue thanks for adding the new `--tileSize` / `-L` option. My previous long running tests do work now. One small hickup remains. Providing tile size via `-L 20` fails for me, but `--tileSize 20` does work:
```
(base) t490s x86_64 /Daten/Git/jlab/gapc_backward>./out ugcauga -p 0:0:0:1  -L 20
Exception: Missing argument of L
(base) t490s x86_64 /Daten/Git/jlab/gapc_backward>./out ugcauga -p 0:0:0:1  --tileSize 20
Checkpointing routine has been integrated. A new checkpoint will be created every 0 days, 0 hours, 0 minutes and 1 seconds.
The checkpoints will be saved at "/Daten/Git/jlab/gapc_backward".

Answer: 
0.123288
```

rtlib/generic_opts.hh
```diff
@@ -198,7 +200,10 @@ class Opts {
         << "                                      after the program finished "
         << "its calculations\n"
 #endif
-       << "\n"
+        << "--tileSize,-L            N            set tile size in "
+        << "multithreaded cyk \n"
+        << "                                      loops (default: 32)\n"
+        << "\n"
```
sjanssen2, line 206: can we hide this parameter if user did not compile with `--cyk` ?

PR #215, opened 2023-06-07
try to reproduce my issue in github actions

sjanssen2:


sjanssen2:
Hey @fymue I found a way to make your test fail: set the interval of archiving such that it occurs before the program finishes, i.e. provide argument `-p 0:0:0:1` to the compiled binary. Did you ever set the interval to such a short time in your tests?

sjanssen2:
the differences between the "correct" output and those with "forgotten output" are not deterministic, i.e. each execution of the binary produces different errors!

fymue:
> Hey @fymue I found a way to make your test fail: set the interval of archiving such that it occurs before the program finishes, i.e. provide argument `-p 0:0:0:1` to the compiled binary. Did you ever set the interval to such a short time in your tests? 

Yes. The tests also didn't fail for such short intervals.

> Vice versa, my tests pass if the interval is the default, i.e. 1h. I feel it might have something to do with switching the active thread?!

That would kind of make sense. I don't know how that could directly mess with the printed output since all checkpointing-related messages are printed to `std::cerr` and the regular program output is printed to `std::cout`, but perhaps Ubuntu can't handle the simultaneous printing to both streams, which theoretically could occur. I will look into it.

Another thing I noticed: The MacOS tests don't even start, because MacOS doesn't seem to know the `-fopenmp` flag. Didn't you successfully run the MacOS tests in the other PR? If so, is there a way you could quickly fix the MacOS OpenMP command so we can see if/how those tests run?

sjanssen2:
> Another thing I noticed: The MacOS tests don't even start, because MacOS doesn't seem to know the `-fopenmp` flag. Didn't you successfully run the MacOS tests in the other PR? If so, is there a way you could quickly fix the MacOS OpenMP command so we can see if/how those tests run?

we use these compile flags for OSX to activate openMP, it should work here as well. I've edited the action file: https://github.com/jlab/gapc/blob/f193b71626d1a4e4f3776294faed60f6ab30b4c2/testdata/regresstest/config#L526-L529

sjanssen2:
If I comment out the instructions to print the `Info: Archived `.... to `std::cerr` in the generated `out.hh` file, I no longer see the discrepancies in the outputs?! Looks like somehow std::cerr and std::cout interfere with each other, but why?

fymue:
> If I comment out the instructions to print the `Info: Archived `.... to `std::cerr` in the generated `out.hh` file, I no longer see the discrepancies in the outputs?! Looks like somehow std::cerr and std::cout interfere with each other, but why?

Alright, that confirms that the simultaneous access to the streams is the issue. At least we figured that out. 

I tried to figure out how/why this could cause that some output disappears/doesn't get added to the stream, but haven't been able to pinpoint the issue. It shouldn't be a buffering issue (meaning that certain lines are buffered but never actually printed), since the missing lines appear at random places in the output and not at the end of the program. Since `stdout` and `stderr` are 2 different streams, they obviously can't get mixed up in any way either. 

My best guess is that the kernel has problems handling the simultaneous access to both streams from two processes that belong to the same parent process, perhaps in combination with the redirection of the output to a file. The kernel can obviously sync access to streams for different processes. In this case, however, the parent process (aka. the program) spawns a child process (aka. the archiving thread), which both access different streams, but simultaneously. Maybe that can cause confusion, because the child and parent process aren't completely separate processes and there are maybe limitations to the ability to sync simultaneous access to 2 streams in that situation? To find the root of this issue, we would probably need to know way more about how File I/O is handled in the kernel.

One way we could fix this is here is to lock a mutex before printing to `std::cout` in `report_insideoutside` and before printing to `std::cerr` in the `archive` methods. This would ensure that each thread would have to wait a tiny amount before printing sth. if the mutex is currently locked (which will never be for long; this won't cause performance issues) and should probably resolve this issue.

This is still a really weird issue though. Especially since it doesn't seem to appear on MacOS, meaning it probably really is a Linux kernel issue, at least for the Ubuntu kernel versions we've used.

fymue:
I was also just able to reproduce the issue on my system by simply dumping a checkpoint every second for a large input. In my previous tests I always either killed the program after the first dump (i.e. after 1s) and then reloaded the program from that checkpoint, or didn't dump that frequently, which didn't cause any problems since fewer messages were printed to `stderr`. So the problem really is caused by repeated logging to `stderr` while also logging to `stdout`. 

This can easily be fixed with a mutex, which I already tested. Do you want me to implement this into the compiler? Since we only every print something at the end of a program to display the result, this should only affect the `print_...` and `report_insideoutside` functions.

sjanssen2:
I would grearly appreciate a mutex solution :-) Thanks!

Fynn Mürmanns ***@***.***> schrieb am So., 11. Juni 2023,
17:09:

PR #218, opened 2023-06-19
escape CHAR('&') for plot-grammar

sjanssen2:
Properly escape `CHAR('&')` via `CHAR('&amp;')` for graphviz output to avoid e.g.
image

### DIFF
--- a/src/fn_arg.hh
+++ b/src/fn_arg.hh
@@ -136,6 +136,8 @@ class Base {
 
     virtual void print(std::ostream &s) = 0;
 
+    void to_dot(std::ostream &out);
+
     virtual bool choice_set() = 0;
 
  protected:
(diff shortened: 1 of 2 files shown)


sjanssen2:
> Do we need something like this for other characters as well? What about `<` or `>`?

@kmaibach you are probably right! Question remains, which characters should be escaped? I started from here https://stackoverflow.com/questions/5665231/most-efficient-way-to-escape-xml-html-in-c-string and found that `'` and `"` work natively, thus it is `>`, `<` and `&` (for now?!)

kmaibach:
> > Do we need something like this for other characters as well? What about `<` or `>`?
> 
> @kmaibach you are probably right! Question remains, which characters should be escaped? I started from here https://stackoverflow.com/questions/5665231/most-efficient-way-to-escape-xml-html-in-c-string and found that `'` and `"` work natively, thus it is `>`, `<` and `&` (for now?!)

Yes, that's what I found as well. I believe it's good for now. We can just extend this should other issues arise

kmaibach (approved):

PR #221, opened 2023-12-14
Tikz

sjanssen2:
- [x] revert gapc-test-suite branch back to master before merging!
- [x] replace `_` with `\_` in algebra function names
- [x] escape further LaTeX special characters for terminal parsers
- [x] let user know about empty result list
- [x] find a way to distinguish between RNA Subsequence and "normal" Subsequence, e.g. adpf.gap VS flowgram2b.gap
- [x] represent raw SEQ as chars, not as <i,j> (similar as rna append_depp_...)
- [ ] warn user if tikz output exceeds latex limits, see https://tex.stackexchange.com/questions/695000/tex-capacity-exceeded-sorry-grouping-levels-255
- [x] print nt parameters
- [x] print LaTeX code for `--kbacktracing`
- [x] print LaTeX code for `--backtracing`
- [x] print LaTeX code for "subopt"
- [x] stochastic backtracing
- [x] window mode
- [ ] find a way to layout tree without overlaps
- [x] Changelog


src/parser.y
```diff
@@ -750,12 +750,12 @@ algebra: algebra_head '{' fn_defs '}'
...
                error(driver.ast.algebras[*$2]->location, "here.");
              } else {
-               Algebra *algebra = driver.ast.signature->generate($2, $4);
+               Algebra *algebra = driver.ast.signature->generate($2, $4, driver.ast.input);
                if (algebra)
                  driver.ast.algebras[*$2] = algebra;
                else
                  error(@4, "Unknown automatic modifier " + *$4 +
-                       ". Use something like count ...");
+                       ". Use 'count', 'enum' or 'tikz'.");
```
kmaibach, line 758: This is so much better than the version before!

rtlib/rope.hh
```diff
@@ -754,6 +756,49 @@ inline void append(rope::Ref<X> &str, const T &x) {
...
+   * LaTeX's "normal" text mode. */
+  boost::replace_all(out, "\\", "\\backslash");
+  boost::replace_all(out, "#", "\\#");
+  boost::replace_all(out, "$", "\\$");
+  boost::replace_all(out, "%", "\\%");
+  boost::replace_all(out, "&", "\\&");
+  boost::replace_all(out, "_", "\\_");
+  boost::replace_all(out, "{", "\\{");
+  boost::replace_all(out, "}", "\\}");
+  boost::replace_all(out, "^", "\\hat{\\ }");
```
kmaibach, line 776: as I just learned the symbol `^` is called `caret` instead of hat :D

rtlib/rna.hh
```diff
@@ -191,6 +191,20 @@ inline void append_deep_rna(
     str.append(static_cast<char>(base_to_char(*i)));
 }
 
+// same as append_deep_rna, but LOC will leave a <X> instead of an empty string
```
kmaibach, line 194: Is <X> the best way to represent this in tikz? Or is it a requirement? Would <EMPTY> not be better for understanding?

kmaibach: also a short comment to `append_deep_rna` about what that does might be useful

kmaibach (changes requested):

rtlib/rope.hh
sjanssen2: hm, it looks like TeX sticks to hat and does not render properly with `\caret`. Who is wrong ;-)

rtlib/rna.hh
sjanssen2: could be interesting for the user to know, where exactly LOC is located, e.g. for dangling bases in RNA problems in algebra functions like [edlr](https://github.com/jlab/fold-grammars/blob/26a959892938f4d42cca8fe73cec35009b7b9046/Grammars/gra_microstate.gap#L15)

PR #240, opened 2025-02-17
WIP: Adding Sycl Support to CYK Calculations

crsct:
This is the pull request that should be used to discuss the sycl specific changes. It is currently only a draft.

### DIFF
--- a/src/cpp.cc
+++ b/src/cpp.cc
@@ -105,6 +105,51 @@ void Printer::Cpp::print(const Statement::For &stmt) {
   stream << stmt.statements;
 }
 
+void Printer::Cpp::print(const Statement::SYCL_Host_Accessor_Decl &stmt) {
+  assert(stmt.name);
+
+  stream << indent() << "sycl::host_accessor " << stmt.name->name << "_hacc{" << stmt.name->name << "};" << endl;
+}
+
+
+void Printer::Cpp::print(const Statement::SYCL_Buffer_Decl &stmt) {
+  assert(stmt.type);
+  assert(stmt.dimension);
+  assert(stmt.name);
+  assert(stmt.value);
+
+  stream << indent() << 
+    "sycl::buffer<" << *stmt.type << "," << stmt.dimension << "> " <<
+    *stmt.name << "(sycl::range<" << stmt.dimension << ">" << 
+    "(" << *stmt.value->name << "));" << endl;
+}
+
+void Printer::Cpp::print(const Statement::SYCL_Accessor_Decl &stmt) {
+  assert(stmt.variable);
+  assert(stmt.conext);
+  
+  stream << indent() << "auto " << *stmt.variable->name + "_acc = sycl::accessor{"
+  << *stmt.variable->name << ", " << *stmt.context->name << ", ";
+  if (*stmt.write && *stmt.read) {
+    stream << "sycl::access_mode::read_write";
+  } else if (*stmt.write) {
+    stream << "sycl::access_mode::write";
+  } else {
+    stream << "sycl::access_mode::read";
+  }
+  stream << "};" << endl; 
+}
+
+void Printer::Cpp::print(const Statement::SYCL_Submit_Kernel &stmt) {
+  assert(stmt.queue);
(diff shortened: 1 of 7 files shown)


src/cpp.cc
```diff
@@ -21,48 +21,44 @@
...
+#include "expr.hh"
+#include "expr/new.hh"
+#include "fn_def.hh"
+#include "grammar.hh"
+#include "options.hh"
+#include "outside/codegen.hh"
 #include "statement.hh"
 #include "statement/backtrace_decl.hh"
+#include "statement/fn_call.hh"
 #include "statement/hash_decl.hh"
```
sjanssen2: This is a bit confusing. Why did you reorder the imports? Current tests break because of some import issues, is this here the cause? Hard to debug. Could you please revert as many lines as possible to show what really changed?

src/cpp.cc
```diff
@@ -21,48 +21,44 @@
...
 
-
 static std::string make_comments(const std::string &s, const std::string &c) {
   std::ostringstream o;
   boost::char_separator<char> nl("\n", "", boost::keep_empty_tokens);
   boost::tokenizer<boost::char_separator<char> > lines(s, nl);
   for (boost::tokenizer<boost::char_separator<char> >::iterator i =
-       lines.begin(); i != lines.end(); ++i) {
+           lines.begin();
+       i != lines.end(); ++i) {
```
sjanssen2: why this change?
If you care about code style - and I am a fan of this - we might want to split your PR into one that adds functionality and a second one only addressing code style changes

src/cpp.hh
```diff
@@ -21,225 +21,227 @@
...
 #include "codegen.hh"
-
-#include "table.hh"
-#include "symbol_fwd.hh"
+#include "operator.hh"
 #include "para_decl_fwd.hh"
+#include "printer.hh"
 #include "product.hh"
+#include "symbol_fwd.hh"
+#include "table.hh"
```
sjanssen2: same critique applies here as for `src/cpp.cc`. This PR is overly hard to understand as the many thousand lines of code that changed just because of code-style are very hard to differentiate from conceptional changes :-/

sjanssen2 (changes requested):
Phu, I simply cannot do the review as too many changes are shown that are irrelevant to your work. Could you please a) revert all code style dependent changes and or b) separate them into an independent PR?

src/cpp.cc
```diff
@@ -105,6 +105,59 @@ void Printer::Cpp::print(const Statement::For &stmt) {
...
+
+  stream << indent() <<
+    "sycl::buffer<" << *stmt.type << "," << stmt.dimension << "> " <<
+    *stmt.name << "(sycl::range<" << stmt.dimension << ">" <<
+    "(" << *stmt.value->name << "));" << endl;
+}
+
+void Printer::Cpp::print(const Statement::SYCL_Accessor_Decl &stmt) {
+  assert(stmt.variable);
+  assert(stmt.conext);
```
sjanssen2, line 132: typo: context !?

src/cyk.cc
```diff
@@ -1299,7 +1300,28 @@ Fn_Def *print_CYK(const AST &ast) {
...
     }
-    fn_cyk->stmts.push_back(new Statement::CustomCode("#pragma omp parallel"));
+
+    int dimension = 1;
+    std::string name = "test";
+    Statement::Var_Decl  *value = new Statement::Var_Decl(
+      new Type::String, "test_value");
+
+    fn_cyk->stmts.push_back(
+      new Statement::SYCL_Buffer_Decl(new Type::Int, dimension, value, value));
```
sjanssen2, line 1310: if you are going to use such a statement, don't you have to include according cycl/cycl.hpp header files?! The currently failing tests point to exactly these headers being absent.

sjanssen2: We furthermore might want to think about a user switch/flag to decide if generated code shall contain sycl "extensions" or not?!

crsct: yes, this is the plan for this week, implementing headers and the proper loop

PR #243, opened 2025-11-04
Smaller Fixes

User-Domme:
-  Re-enabled C++17 support (previously forced C++14 for legacy reasons).
- Fixed sed expression that mistakenly removed '-l' inside terms like '-linux'.
- Added PyTorch include/library path detection for Conda environments.

### DIFF
--- a/src/cpp.cc
+++ b/src/cpp.cc
@@ -3346,8 +3346,6 @@ void Printer::Cpp::pytorch_makefile(const Options &opts, const AST &ast) {
     stream << "CXXFLAGS += -DMALLOC_BATCH" << endl;
     stream << "endif" << endl << endl;
   }
-  // pytorch C++ extension doesn't support C++17 yet (but not nessecary anymore)
-//  stream << "CXXFLAGS := $(CXXFLAGS) | sed 's/c++17/c++14/'" << endl;
 
   stream << "compiler_args := $(CXXFLAGS) | sed -e 's/\\s/\", \"/g'" << endl;
   stream << "LDFLAGS := $(LDFLAGS) | sed 's/-L\\//\\//g'" << endl;
--- a/src/cpp.cc
+++ b/src/cpp.cc
@@ -3346,8 +3346,8 @@ void Printer::Cpp::pytorch_makefile(const Options &opts, const AST &ast) {
     stream << "CXXFLAGS += -DMALLOC_BATCH" << endl;
     stream << "endif" << endl << endl;
   }
-  // pytorch C++ extension doesn't support C++17 yet
-  stream << "CXXFLAGS := $(CXXFLAGS) | sed 's/c++17/c++14/'" << endl;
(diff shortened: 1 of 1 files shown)


src/cpp.cc
```diff
@@ -3346,8 +3346,8 @@ void Printer::Cpp::pytorch_makefile(const Options &opts, const AST &ast) {
     stream << "CXXFLAGS += -DMALLOC_BATCH" << endl;
     stream << "endif" << endl << endl;
   }
-  // pytorch C++ extension doesn't support C++17 yet
-  stream << "CXXFLAGS := $(CXXFLAGS) | sed 's/c++17/c++14/'" << endl;
+  // pytorch C++ extension doesn't support C++17 yet (but not nessecary anymore)
+//  stream << "CXXFLAGS := $(CXXFLAGS) | sed 's/c++17/c++14/'" << endl;
```
sjanssen2: I'd simply remove those two lines

src/cpp.cc
```diff
@@ -3357,7 +3357,7 @@ void Printer::Cpp::pytorch_makefile(const Options &opts, const AST &ast) {
          << "sed -e 's/\\s/\", \"-Xlinker\", \"-rpath\", \"-Xlinker\", \"/g'"
          << endl;
   stream << "library_dirs := $(LDFLAGS) | sed -e 's/\\s/\", \"/g'" << endl;
-  stream << "LDLIBS := $(LDLIBS) | sed 's/-l//g'" << endl;
+  stream << "LDLIBS := $(LDLIBS) | sed -E 's/(^|[[:space:]])-l/\\1/g'" << endl;
```
sjanssen2, line 3358: not sure if this is relevant here, but sed behaves differently in OSX and Linux
Can you also show an example of what sed should change here?
what does the `|` mean in the regex? why are you capturing multiple spaces?

sjanssen2 (changes requested):
Thanks a lot. However, I do have comments / questions

src/cpp.cc
User-Domme: The old sed would falsely remove all "-l" within a term for example:
"-l/usr/lib/x86_64-linux-gnu, -lsd"  --> "/usr/lib/x86_64inux-gnu, sd" the "64-linux" turns into "64inux" which causes an error.

The new expression makes sure the "-l" is either at the start of a line (^) or (|) behind a space, line break or tab ( [[:SPACE:]] ). The "\\1" is needed to keep the capture group that is detected within the "( )" and not fully replace it, otherwise the space between terms would be removed as well and 2 terms would falsely merge into one.

So finally the new expression correctly performs the following:
"-l/usr/lib/x86_64-linux-gnu, -lsd"  --> "/usr/lib/x86_64-linux-gnu, sd"

For OSX the regex should do the same thing, at least i tested it on an OSX terminal as well.

PR #246, opened 2026-05-06
replace boost/cstdint.hpp with csdtint

sjanssen2:


sjanssen2:
Closing in favor of #247

